PR #1, opened 2016-06-26
Expose SignatureSchemes in ClientHelloInfo

grittygrease:
Make the signatureAndHash structure public and expose client supported
signature and has pairs in ClientHelloInfo as SignatureSchemes for use
in certificate selection.


FiloSottile:
Done in dc1c6ad71081994b0fc539312f153f68e76e09ba

PR #4, opened 2016-06-30
Marshal clientHello into conn

grittygrease:


Bren2010:
Thank you for figuring this out. I'm really sorry if I lost it in rebase-machinations.

PR #5, opened 2016-07-01
Adding temporary patch for NSS DHE share bug

martinthomson:
Untested.  YMMV.


martinthomson:
OK, not working, need to brush up on docker basics.

PR #7, opened 2016-07-14
Enable CHACHA20-POLY1305 cipher suites

vkrasnov:


FiloSottile:
I don't think the chacha20 package belongs in tris, or at least the review doesn't.

Let's merge cloudflare/go#14 first, then I'll figure out the best way to pull it in here. It might just be documenting that tris builds only with our compiler for now.

FiloSottile:
This was pulled in piecemeal.

PR #14, opened 2016-12-28
TLS 1.3 short header experiment

ekr:
@FiloSottile I don't think this is ready to merge yet, and I note that it won't pass CI with the short headers tests, but I think it's ready for you to take a first look

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -183,6 +183,10 @@ type halfConn struct {
 	// used to save allocating a new buffer for each MAC.
 	inDigestBuf, outDigestBuf []byte
 
+	// The length of the record header. Always 5 for TLS < 1.2 but maybe
+	// 2 in the TLS 1.3 short header variant.
+	recordHeaderLen int
+	
 	traceErr func(error)
 }
 
@@ -310,7 +314,7 @@ type cbcMode interface {
 // order to get the application payload, and an optional alert value.
 func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert) {
 	// pull out payload
-	payload := b.data[recordHeaderLen:]
+	payload := b.data[hc.recordHeaderLen:]
 
 	macSize := 0
 	if hc.mac != nil {
@@ -352,7 +356,7 @@ func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert)
 			if err != nil {
 				return false, 0, alertBadRecordMAC
 			}
-			b.resize(recordHeaderLen + explicitIVLen + len(payload))
+			b.resize(hc.recordHeaderLen + explicitIVLen + len(payload))
 		case cbcMode:
 			blockSize := c.BlockSize()
 			if hc.version >= VersionTLS11 {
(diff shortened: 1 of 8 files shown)


ekr:
@FiloSottile: OK, as I kind of suspected, I am not setting c.[in,out].recordHeaderLen in the right place, so it's 0 during this failing test. Would appreciate your guidance on how to set it in whatever passes for a constructor.

conn.go
```diff
@@ -663,19 +672,34 @@ func (c *Conn) readRecord(want recordType) error {
...
-	vers := uint16(b.data[1])<<8 | uint16(b.data[2])
-	n := int(b.data[3])<<8 | int(b.data[4])
+		vers = uint16(b.data[1])<<8 | uint16(b.data[2])
+		n = int(b.data[3])<<8 | int(b.data[4])
+	} else {
+		typ = recordTypeApplicationData
+		vers = 0x0301
+		if int(b.data[0]) & 0x80 == 0 {
+			c.sendAlert(alertDecodeError)
+			return c.in.setErrorLocked(c.newRecordHeaderError("Short header high bit not set"))
```
FiloSottile: Errors start lowercase

conn.go
```diff
@@ -183,6 +183,10 @@ type halfConn struct {
 	// used to save allocating a new buffer for each MAC.
 	inDigestBuf, outDigestBuf []byte
 
+	// The length of the record header. Always 5 for TLS < 1.2 but maybe
+	// 2 in the TLS 1.3 short header variant.
+	recordHeaderLen int
```
FiloSottile: Go doesn't have initializators, instead the philosophy is to make the zero value useful.

In this case you would have a bool flag here instead of an int, and then do some

```
recordHeaderLen := normalRecordHeaderLen
if hc.shortRecordHeader {
    recordHeaderLen = shortRecordHeaderLen
}
```

If you find yourself writing that snippet more than 2-3 times, and you really hate it, make a `recordHeaderLen() int` method on `halfConn`.

It might seem less flexible, but you already have `== normalRecordHeaderLen` somewhere, so that flexibility is an illusion and it's best removed. (There are only two valid states of this int, it might as well be a bool.)

conn.go
```diff
@@ -429,37 +433,37 @@ func padToBlockSize(payload []byte, blockSize int) (prefix, finalBlock []byte) {
 }
 
 // encrypt encrypts and macs the data in b.
-func (hc *halfConn) encrypt(b *block, explicitIVLen int) (bool, alert) {
+func (hc *halfConn) encrypt(t byte, b *block, explicitIVLen int) (bool, alert) {
```
FiloSottile: Note to self: this function got out of control, the abstraction is fully broken, it might as well make the header itself. For future refactor.

handshake_messages.go
```diff
@@ -898,7 +913,6 @@ func (m *serverHelloMsg) unmarshal(data []byte) bool {
 	m.scts = nil
 	m.ticketSupported = false
 	m.alpnProtocol = ""
-
```
ekr: Just spurious, I guess.

conn.go
ekr: Acknowledged.

conn.go
ekr: ack.

ekr:
I did a Google CLA for libjingle, so we should be covered.

FiloSottile:
You need to add a randomized assignment here for the fuzzing test to be useful https://github.com/cloudflare/tls-tris/blob/master/handshake_messages_test.go#L206

FiloSottile:
Functions that use the recordHeaderLen a lot should just assign a local variable at the top instead of calling it 10 times.

If there are few enough functions using it, you could also just inline it.

handshake_messages_test.go
```diff
@@ -210,6 +210,7 @@ func (*serverHelloMsg13) Generate(rand *rand.Rand, size int) reflect.Value {
 	m.cipherSuite = uint16(rand.Int31())
 	m.keyShare.group = CurveID(rand.Intn(30000))
 	m.keyShare.data = randomBytes(rand.Intn(300), rand)
+	m.shortHeaders.data = bool(rand)
```
FiloSottile: Does... that actually compile/work? Everything else uses `if rand.Intn(10) > 5 {`

FiloSottile: It should be set also in `(*clientHelloMsg).Generate`

ekr:
@bifurcation, @FiloSottile: this is getting a mismatch on secureRenegotiation in the Marshal/Unmarshal test. Do you see anythign obvious?

Lekensteyn:
Chrome removed this experiment in Mar 2017, it should probably be removed too in tls-tris. (Revert has some trivial conflicts that needs to be solved).
https://bugs.chromium.org/p/boringssl/issues/detail?id=119

PR #16, opened 2017-01-24
Support OCSP stapling.

Bren2010:


handshake_messages_test.go
```diff
@@ -252,9 +252,9 @@ func (*certificateMsg) Generate(rand *rand.Rand, size int) reflect.Value {
 func (*certificateMsg13) Generate(rand *rand.Rand, size int) reflect.Value {
 	m := &certificateMsg13{}
 	numCerts := rand.Intn(20)
-	m.certificates = make([][]byte, numCerts)
+	m.certificates = make([]certificateEntry, numCerts)
 	for i := 0; i < numCerts; i++ {
-		m.certificates[i] = randomBytes(rand.Intn(10)+1, rand)
+		m.certificates[i].data = randomBytes(rand.Intn(10)+1, rand)
```
FiloSottile: Add conditional fuzzing (sometimes nil sometimes not) for both SCT and OCSP

handshake_messages.go
```diff
@@ -1407,10 +1407,16 @@ func (m *certificateMsg) unmarshal(data []byte) alert {
 	return alertSuccess
 }
 
+type certificateEntry struct {
+	data       []byte
+	ocspStaple []byte
+	sctList    []byte
```
FiloSottile: Since this is implemented, can you add SignedCertificateTimestamps support above while you are at it?

handshake_messages.go
```diff
@@ -1455,12 +1478,47 @@ func (m *certificateMsg13) marshal() (x []byte) {
...
-		copy(z[3:], slice)
-		z = z[3+len(slice)+2:]
+	for _, cert := range m.certificates {
+		z[0] = uint8(len(cert.data) >> 16)
+		z[1] = uint8(len(cert.data) >> 8)
+		z[2] = uint8(len(cert.data))
+		copy(z[3:], cert.data)
+		z = z[3+len(cert.data):]
+
+		temp := z[:2]
```
FiloSottile: `temp` means as much as `x`. So use `x` or better `extLenPos`.

handshake_messages.go
```diff
@@ -1407,10 +1407,16 @@ func (m *certificateMsg) unmarshal(data []byte) alert {
 	return alertSuccess
 }
 
+type certificateEntry struct {
```
FiloSottile: These are all in common.go.

PR #21, opened 2017-03-24
Fix Content-Type header for /ch endpoint

tmthrgd:
It's a really minor thing that has no bearing on anything, but, for the sake of correctness, the `/ch` endpoint in tris-localserver sets the `Content-Type` header on the `*http.Request` rather than on the `http.ResponseWriter`.

This PR makes no other changes. I have signed the Google CLA for what it is worth.


FiloSottile:
Ah, yep, thanks ^^

PR #22, opened 2017-03-26
Proposal: remove TLS13CipherSuites from tls.Config

tmthrgd:
This is a proposal to eliminate the `TLS13CipherSuites` field from `Config` without changing current behaviour or functionality.

My understanding is that `TLS13CipherSuites` was added to allow servers that explicitly configure `CipherSuites` to automatically gain TLS 1.3 support without requiring any code changes. This proposal addresses this problem **without** requiring a separate field for TLS 1.3 cipher suites - which is IMO very messy (and `Config` really doesn't need to be made worse)!

Instead of maintaining separate lists of TLS 1.0-1.2 and TLS 1.3 cipher suites, the logic of `(*Config).cipherSuites` is modified when negotiating TLS 1.3. Iff `CipherSuites` does not contain a supported TLS 1.3 cipher suite, the default cipher suite list is used.

Note: This currently breaks client cipher suite advertisement if version is TLS 1.3, but I believe it was already broken and this can be fixed later if this proposal is accepted.

This is related to: https://go-review.googlesource.com/c/33419/

### DIFF
--- a/common.go
+++ b/common.go
@@ -505,14 +505,11 @@ type Config struct {
 	// This should be used only for testing.
 	InsecureSkipVerify bool
 
-	// CipherSuites is a list of supported cipher suites to be used in
-	// TLS 1.0-1.2. If CipherSuites is nil, TLS uses a list of suites
-	// supported by the implementation.
+	// CipherSuites is a list of supported cipher suites. If CipherSuites
+	// is nil, TLS uses a list of suites supported by the implementation.
+	// If CipherSuites does not contain a TLS1.3 suite, TLS uses a list
+	// of suites supported by the implementation when negotiating TLS1.3.
 	CipherSuites []uint16
-
-	// TLS13CipherSuites is a list of supported cipher suites to be used in
-	// TLS 1.3. If nil, uses a list of suites supported by the implementation.
-	TLS13CipherSuites []uint16
 
 	// PreferServerCipherSuites controls whether the server selects the
 	// client's most preferred ciphersuite, or the server's most preferred
@@ -659,7 +656,6 @@ func (c *Config) Clone() *Config {
 		ClientCAs:                   c.ClientCAs,
 		InsecureSkipVerify:          c.InsecureSkipVerify,
 		CipherSuites:                c.CipherSuites,
-		TLS13CipherSuites:           c.TLS13CipherSuites,
 		PreferServerCipherSuites:    c.PreferServerCipherSuites,
 		SessionTicketsDisabled:      c.SessionTicketsDisabled,
 		SessionTicketKey:            c.SessionTicketKey,
@@ -758,11 +754,20 @@ func (c *Config) time() time.Time {
 
 func (c *Config) cipherSuites(version uint16) []uint16 {
 	if version >= VersionTLS13 {
-		s := c.TLS13CipherSuites
-		if s == nil {
-			s = defaultTLS13CipherSuites()
+	NextCipherSuite:
+		for _, id := range c.CipherSuites {
+			for _, s := range cipherSuites {
+				if s.id != id {
+					continue
+				}
+				if s.flags&suiteTLS13 == 0 {
+					continue NextCipherSuite
+				}
+				return c.CipherSuites
+			}
 		}
-		return s
+		// TODO: this breaks client cipher advertisement.
+		return defaultCipherSuites()
 	}
 	s := c.CipherSuites
 	if s == nil {
@@ -1037,9 +1042,8 @@ func defaultConfig() *Config {
 }
 
 var (
-	once                        sync.Once
-	varDefaultCipherSuites      []uint16
-	varDefaultTLS13CipherSuites []uint16
+	once                   sync.Once
+	varDefaultCipherSuites []uint16
 )
 
 func defaultCipherSuites() []uint16 {
@@ -1047,22 +1051,15 @@ func defaultCipherSuites() []uint16 {
 	return varDefaultCipherSuites
 }
 
-func defaultTLS13CipherSuites() []uint16 {
-	once.Do(initDefaultCipherSuites)
-	return varDefaultTLS13CipherSuites
-}
-
 func initDefaultCipherSuites() {
-	var topCipherSuites, topTLS13CipherSuites []uint16
+	var topCipherSuites []uint16
 	if cipherhw.AESGCMSupport() {
 		// If AES-GCM hardware is provided then prioritise AES-GCM
 		// cipher suites.
-		topTLS13CipherSuites = []uint16{
+		topCipherSuites = []uint16{
 			TLS_AES_128_GCM_SHA256,
 			TLS_AES_256_GCM_SHA384,
 			TLS_CHACHA20_POLY1305_SHA256,
-		}
-		topCipherSuites = []uint16{
 			TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 			TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
 			TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
@@ -1073,12 +1070,10 @@ func initDefaultCipherSuites() {
 	} else {
 		// Without AES-GCM hardware, we put the ChaCha20-Poly1305
 		// cipher suites first.
-		topTLS13CipherSuites = []uint16{
+		topCipherSuites = []uint16{
 			TLS_CHACHA20_POLY1305_SHA256,
 			TLS_AES_128_GCM_SHA256,
 			TLS_AES_256_GCM_SHA384,
-		}
-		topCipherSuites = []uint16{
 			TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
 			TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
 			TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
@@ -1088,10 +1083,6 @@ func initDefaultCipherSuites() {
 		}
 	}
 
-	varDefaultTLS13CipherSuites = make([]uint16, 0, len(cipherSuites))
-	for _, topCipher := range topTLS13CipherSuites {
-		varDefaultTLS13CipherSuites = append(varDefaultTLS13CipherSuites, topCipher)
-	}
 	varDefaultCipherSuites = make([]uint16, 0, len(cipherSuites))
 	varDefaultCipherSuites = append(varDefaultCipherSuites, topCipherSuites...)
 
@@ -1100,21 +1091,12 @@ NextCipherSuite:
 		if suite.flags&suiteDefaultOff != 0 {
 			continue
 		}
-		if suite.flags&suiteTLS13 != 0 {
-			for _, existing := range varDefaultTLS13CipherSuites {
-				if existing == suite.id {
-					continue NextCipherSuite
-				}
+		for _, existing := range varDefaultCipherSuites {
+			if existing == suite.id {
+				continue NextCipherSuite
 			}
-			varDefaultTLS13CipherSuites = append(varDefaultTLS13CipherSuites, suite.id)
-		} else {
-			for _, existing := range varDefaultCipherSuites {
-				if existing == suite.id {
-					continue NextCipherSuite
-				}
-			}
-			varDefaultCipherSuites = append(varDefaultCipherSuites, suite.id)
 		}
+		varDefaultCipherSuites = append(varDefaultCipherSuites, suite.id)
 	}
 }
 
--- a/handshake_client.go
+++ b/handshake_client.go
@@ -86,7 +86,12 @@ NextCipherSuite:
 			// Don't advertise TLS 1.2-only cipher suites unless
 			// we're attempting TLS 1.2.
 			if hello.vers < VersionTLS12 && suite.flags&suiteTLS12 != 0 {
-				continue
+				continue NextCipherSuite
+			}
+			// Don't advertise TLS 1.3-only cipher suites unless
+			// we're attempting TLS 1.3.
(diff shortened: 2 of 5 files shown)


handshake_client.go
```diff
@@ -86,7 +86,12 @@ NextCipherSuite:
 			// Don't advertise TLS 1.2-only cipher suites unless
 			// we're attempting TLS 1.2.
 			if hello.vers < VersionTLS12 && suite.flags&suiteTLS12 != 0 {
-				continue
+				continue NextCipherSuite
+			}
+			// Don't advertise TLS 1.3-only cipher suites unless
+			// we're attempting TLS 1.3.
+			if hello.vers < VersionTLS13 && suite.flags&suiteTLS13 != 0 {
+				continue NextCipherSuite
```
Lekensteyn: Shouldn't both of these just be a normal `continue`? Otherwise it will not try to match other potential cipher suites from `cipherSuites` against the proposed candidate from the configuration (`possibleCipherSuites`) once it notices an incompatible cipher suite.

common.go
```diff
@@ -770,11 +766,20 @@ func (c *Config) time() time.Time {
...
+				}
+				if s.flags&suiteTLS13 == 0 {
+					continue NextCipherSuite
+				}
+				return c.CipherSuites
+			}
 		}
-		return s
+		// TODO: this breaks client cipher advertisement.
+		return defaultCipherSuites()
```
Lekensteyn: This is undesirable I think, it ignores the requested CS. I think it is better to reduce maxVersion to 1.2 if no supported TLS 1.3 suite is available. When minVersion is 1.3, then it should become an error. What do you think?

handshake_client.go
tmthrgd: @Lekensteyn I'm pretty sure the logic is correct. It's not hugely important as it's really nothing more than an optimisation.

The outer loop is the configured cipher suites and the inner loop is the crypto/tls supported cipher suites. That's the desired behaviour because cipher suites have unique ids so no more than one suite (from `cipherSuites`) can ever match. (Unless I'm fatally misunderstanding something here).

common.go
tmthrgd: @Lekensteyn This is rough and I only put this in here for the sake of demonstrating the proposed feature. If this was accepted, this would *definitely* be fixed.

(For context: "Note: This currently breaks client cipher suite advertisement if version is TLS 1.3, but I believe it was already broken and this can be fixed later if this proposal is accepted.")

That's not a bad idea, but it does go against the express wishes of @FiloSottile. ("I opted for adding a separate TLS13CipherSuites field to the Config because library users that did not set Config.MaxVersion are supposed to get TLS 1.3 support automatically, like it has been for HTTP/2, but having set CipherSuites would effectively disable it.")

To elaborate on why I think this is already broken, `(*Conn).clientHandshake` needs to send both the TLS 1.2 and bellow **and** the TLS 1.3 cipher suites in the client hello. Currently it exclusively sends TLS 1.3 cipher suites when configured for TLS 1.3 (if I'm reading the logic correctly). This is perfectly fine for `(*serverHandshakeState).readClientHello` where only the current version cipher suites are needed.

The fix that seems to make most sense to me is something like this (again rough):
```
NextCipherSuite:
	for _, id := range c.CipherSuites {
		for _, s := range cipherSuites {
			...
		}
	}
	s := c.CipherSuites
	if s == nil {
		return defaultCipherSuites()
	}
	return append(s, defaultTLS13CipherSuites()...)
```

With the above fix, server cipher selection will still work fine (with slight memory overhead, that could be avoided), and the client will offer the configured TLS 1.2 cipher suites followed by the default TLS 1.3 cipher suites.

An alternative would be to only automatically support TLS 1.3 for the server and **not** the client in this case.

Lekensteyn: > Currently [the client] exclusively sends TLS 1.3 cipher suites when configured for TLS 1.3

Indeed broken (in common.go), but TLS 1.3 client support is still work in progress, so the issue has not surfaced yet.

If the concern is enabling TLS 1.3 by default when the user has overridden CipherSuites, what about this adjusted logic (in common.go):

```go
func (c *Config) cipherSuites(version uint16) []uint16 {
	s := c.CipherSuites
	if s == nil {
		s = defaultCipherSuites()
	}
	if c.MaxVersion == 0 || (version >= VersionTLS13 && c.CipherSuites == nil) {
		s = append(s, defaultTLS13CipherSuites())
	}
	return s
}
```
that does not require inspecting the given cipher suites list. Of course if the user sets CipherSuites without any TLS 1.3 ciphers and MaxVersion 1.3, then the handshake will fail. But likewise, if the user sets CipherSuites with only TLS 1.3 ciphers, MinVersion 1.2 would not be that useful.

Lekensteyn:
@FiloSottile Reasons to make the cipher configurable are mainly for flexibility in testing, not actual client/server applications:
- Testing the cipher configuration (internal implementation). If the codepath with ChaCha20-Poly1305 is never executed except when AES hw accel is unavailable, then it might result in surprises.
- Testing of crypto/tls against external configurations. boringssl currently has the TLS 1.3 cipher suites hardcoded. This seems reasonable given the currently available cipher suites, but again makes testing harder.
- Related to the previous point, forcing a certain cipher might be desirable if you want to test external tools (ChaCha20-Poly1305 pcap in Wireshark, vulnerabilities, etc.).

FiloSottile:
I see those points, but I don't think testing needs justify a public API, and agl made that point before upstream, so if this is targeted for upstreaming I suggest not making it an exported API.

Lekensteyn:
@FiloSottile Could you refer to the upstream discussion? Can't find anything in https://go-review.googlesource.com/q/project:go+branch:dev.tls and the closest I get for "tls 1.3 go cipher suites configurable" is https://bugs.chromium.org/p/boringssl/issues/detail?id=110

David's comment in the above boringssl bug seems reasonable, we could do the same. Unfortunately that would mean that overriding the cipher needs a patched crypto/tls library, but maybe that is acceptable for the few edge cases.

FiloSottile:
I was referring to https://github.com/golang/go/issues/21593#issuecomment-324784628 which is about something else, but makes the point about not adding APIs with only testing use-cases.

tmthrgd:
@FiloSottile I think there is a relatively strong case for allowing them to be configured. One simply being that although all TLS 1.3 *implemented* cipher suites are *currently* secure that may not always be the case.

(Also keep in mind that TLS 1.3 also defines two other cipher suites, `TLS_AES_128_CCM_SHA256` and `TLS_AES_128_CCM_8_SHA256`. Although neither of these are implemented, it's possible they could be at some point in the future).

Secure is also a relative term, for example `TLS_AES_128_GCM_SHA256` could prove quite vulnerable to quantum computing attacks on symmetric ciphers (should they ever materialise) which might make it insufficiently secure for certain uses.

It also doesn't seem unreasonable to force a preference for exclusively 256-bit ciphers (for inter-data-centre communication for example). I know for a fact that Cloudflare already does this with [gokeyless](https://github.com/cloudflare/gokeyless) ([client cipher suites](https://github.com/cloudflare/gokeyless/blob/6eddd9a121e2b2cb2995d8dd5fefa2076dc8958c/client/client.go#L53-L56) and [server cipher cuites](https://github.com/cloudflare/gokeyless/blob/7e94dd24e1ed1660d953cc409a95163e9dbc4b2b/server/server.go#L153-L156)).

Ultimately I think TLS 1.3 cipher suites *should* be configurable, but I still think it would be a mistake to add a separate config field to do that.

common.go
tmthrgd: @Lekensteyn I think it would be best to change the API slightly, `(*Config).cipherSuites` is unexported so there are no compatibility problems. That would allow this to work perfectly without requiring any copying from the `append` call.

I honestly really like the idea that expressly setting `c.MaxVersion` to 1.3 and not-including TLS 1.3 ciphers should result in the handshake-failing.

FiloSottile:
Most of those arguments (except the one about 256-bit security) seem rooted in future possibilities. APIs can always be added, never removed, so I don't think it's time to add one now.

Lekensteyn:
@tmthrgd I have force-pushed master (see #33, includes rebase on go1.9 crypto/tls) and therefore rebased this branch on top of it (with no other changes).

For SIDH we definitely want to force AES256 over AES128, but that could be handled as an internal detail (which looks at the supported_groups extension). If that is good enough, let's go with @FiloSottile's suggestion on hard-coding the TLS 1.3 suites into the supported ciphersuites list (when MaxVersion >= 1.3).

Wrt CCM, these were intended for IoT applications AFAIK and you will likely not see them on the web. If desired, I guess that the CipherSuites configuration can be extended in the future by recognizing TLS 1.3 cipher suites.

Also, unfortunately Github removed some of the review comments, so I have included the original discussion below.
@Lekensteyn
> ```
> common.go
>  		// TODO: this breaks client cipher advertisement.
> +		return defaultCipherSuites()
> ```
> This is undesirable I think, it ignores the requested CS. I think it is better to reduce maxVersion to 1.2 if no supported TLS 1.3 suite is available. When minVersion is 1.3, then it should become an error. What do you think?

@tmthrgd
> @Lekensteyn This is rough and I only put this in here for the sake of demonstrating the proposed feature. If this was accepted, this would *definitely* be fixed.
> 
> (For context: "Note: This currently breaks client cipher suite advertisement if version is TLS 1.3, but I believe it was already broken and this can be fixed later if this proposal is accepted.")
> 
> That's not a bad idea, but it does go against the express wishes of @FiloSottile. ("I opted for adding a separate TLS13CipherSuites field to the Config because library users that did not set Config.MaxVersion are supposed to get TLS 1.3 support automatically, like it has been for HTTP/2, but having set CipherSuites would effectively disable it.")
> 
> To elaborate on why I think this is already broken, `(*Conn).clientHandshake` needs to send both the TLS 1.2 and bellow **and** the TLS 1.3 cipher suites in the client hello. Currently it exclusively sends TLS 1.3 cipher suites when configured for TLS 1.3 (if I'm reading the logic correctly). This is perfectly fine for `(*serverHandshakeState).readClientHello` where only the current version cipher suites are needed.
> 
> The fix that seems to make most sense to me is something like this (again rough):
> ```
> NextCipherSuite:
> 	for _, id := range c.CipherSuites {
> 		for _, s := range cipherSuites {
> 			...
> 		}
> 	}
> 	s := c.CipherSuites
> 	if s == nil {
> 		return defaultCipherSuites()
> 	}
> 	return append(s, defaultTLS13CipherSuites()...)
> ```
> 
> With the above fix, server cipher selection will still work fine (with slight memory overhead, that could be avoided), and the client will offer the configured TLS 1.2 cipher suites followed by the default TLS 1.3 cipher suites.
> 
> An alternative would be to only automatically support TLS 1.3 for the server and **not** the client whenever this case comes up.

@Lekensteyn
> > Currently [the client] exclusively sends TLS 1.3 cipher suites when configured for TLS 1.3
> 
> Indeed broken (in common.go), but TLS 1.3 client support is still work in progress, so the issue has not surfaced yet.
> 
> If the concern is enabling TLS 1.3 by default when the user has overridden CipherSuites, what about this adjusted logic (in common.go):
> 
> ```go
> func (c *Config) cipherSuites(version uint16) []uint16 {
> 	s := c.CipherSuites
> 	if s == nil {
> 		s = defaultCipherSuites()
> 	}
> 	if c.MaxVersion == 0 || (version >= VersionTLS13 && c.CipherSuites == nil) {
> 		s = append(s, defaultTLS13CipherSuites())
> 	}
> 	return s
> }
> ```
> that does not require inspecting the given cipher suites list. Of course if the user sets CipherSuites without any TLS 1.3 ciphers and MaxVersion 1.3, then the handshake will fail. But likewise, if the user sets CipherSuites with only TLS 1.3 ciphers, MinVersion 1.2 would not be that useful.

@tmthrgd
> @Lekensteyn I think it would be best to change the API slightly, `(*Config).cipherSuites` is unexported so there are no compatibility problems. That would allow this to work perfectly without requiring any copying from the `append` call.
> 
> I honestly really like the idea that expressly setting `c.MaxVersion` to 1.3 and not-including TLS 1.3 ciphers should result in the handshake-failing.

Lekensteyn:
There is an issue with the cipherSuites function, currently it just returns either TLS 1.3 or TLS 1.2 suites. It should return both, the server setCipherSuite method will check for the version anyway.

Since version negotiation has not happened yet, `c.vers = 0` in the client handshake method, so `cipherSuites(c.vers)` works only for TLS 1.2 clients. As TLS 1.3 client support is added, this is however no longer sufficient.

I propose to remove the version field from cipherSuites function.

Lekensteyn:
After some thought, I have decided to make it configurable. See the commit message of https://github.com/cloudflare/tls-tris/commit/3063195bc9e160fc5c10bf7cf4a19025bc3cd2c3

More extended reasoning:
in TLS 1.2 there is a plethora of cipher suites including terribly bad ones. With other libraries, applications would therefore be better off by not relying on bad defaults and instead select stronger cipher suites in their applications. Application developers might not always make decisions that will still be secure in the future (see use of RC4 to mitigate CBC weaknesses), so it might be desirable to remove that configurability option for TLS 1.3 ciphers.

There are not many TLS 1.3 ciphers though and all of the supported ciphers are currently excellent choices. I believe that maintaining the TLS 1.2 feature of configurability of cipher suites will not hurt the API nor security of applications.

Restricting the "CipherSuites" option will **require more logic** and may have potentially unexpected behavior (currently it would just ignore those cipher suites without warnings/panic). **There is no "new" API**. The "CipherSuites" mechanism continues to work as application developers expect. And due to the extra check for missing TLS 1.3 cipher suites, it will still **allow older applications to use TLS 1.3** if they have only specified TLS 1.2 cipher suites without modifying MaxVersion.

@FiloSottile What do you think?

Lekensteyn:
This PR has been superseded by #37 which also includes some commits from this patch. Thanks @tmthrgd for bringing it up!

PR #23, opened 2017-03-27
Use unique ticket_age_add for each TLS 1.3 new ticket.

tmthrgd:
According to the specification, `ticket_age_add` should be unique per ticket. The current behaviour is to reuse the same value for each of the `numSessionTickets` issued ticket. This has been confirmed as a bug in tlswg/tls13-spec#913.

This change fixes this bug and does nothing else.

### DIFF
--- a/13.go
+++ b/13.go
@@ -566,15 +566,9 @@ func (hs *serverHandshakeState) sendSessionTicket13() error {
 	resumptionSecret := hkdfExpandLabel(hash, hs.masterSecret, handshakeCtx, "resumption master secret", hash.Size())
 
 	ageAddBuf := make([]byte, 4)
-	if _, err := io.ReadFull(c.config.rand(), ageAddBuf); err != nil {
-		c.sendAlert(alertInternalError)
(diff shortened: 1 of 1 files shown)


tmthrgd:
The required behaviour has been clarified by tlswg/tls13-spec@ee48e213a0aeec2130eca11fd10236869e58dbb7.

FiloSottile:
@tmthrgd Very good catch, thank you!

To accept this PR, I need you to explicitly confirm the following, to allow the patch to be upstreamed into Go:

- [ ] you signed the Google CLA with the email me+github@tomthorogood.co.uk. If you did with a different email, please tell me which one and I'll amend the commit.

- [ ] you are ok with me submitting this patch on your behalf to the Go project, under the terms of the Google CLA

tmthrgd:
@FiloSottile You're welcome. I've signed the Google CLA with <me+google@tomthorogood.co.uk> and I'm perfectly ok with you submitting this upstream on my behalf under those terms.

PR #25, opened 2017-06-09
Initial TLS 1.3 client support

hdevalence:
This branch has initial support for a TLS 1.3 client.

When acting as a client, we send a 1.3/1.2 ClientHello and fall back to the existing TLS 1.2 client code if the server sends a 1.2 Server Hello.

PSK features are not implemented.
Responding to HelloRetryRequests is not implemented.

Cert validation is not implemented -- the client will panic unless the InsecureSkipVerify option is set -- because I felt like it would be better to have that logic reviewed separately.  However, if it's better to do it all at once, I can implement it in this branch before it's merged.


13.go
```diff
@@ -641,3 +641,244 @@ func (hs *serverHandshakeState) traceErr(err error) {
...
+		c.sendAlert(alertUnexpectedMessage)
+		return unexpectedMessageError(encryptedExt, msg)
+	}
+	hs.encryptedExt = encryptedExt
+
+	hs.finishedHash13.Write(encryptedExt.marshal())
+
+	// If we use PSKs, jump to state_WAIT_FINISHED here
+	// XXX what happens if we request to use a PSK and the
+	// server wants to do a full handshake?
```
FiloSottile: The server is free to reject PSK and we just do a full handshake.

key_agreement.go
```diff
@@ -114,10 +114,18 @@ func md5SHA1Hash(slices [][]byte) []byte {
 // only used for >= TLS 1.2 and precisely identifies the hash function to use.
 func hashForServerKeyExchange(sigAndHash signatureAndHash, version uint16, slices ...[]byte) ([]byte, crypto.Hash, error) {
 	if version >= VersionTLS12 {
-		if !isSupportedSignatureAndHash(sigAndHash, supportedSignatureAlgorithms) {
+		var hashFunc crypto.Hash
+		var err error
+		if signatureSchemeIsPSS(sigAndHashToSigScheme(sigAndHash)) {
+			// 1.3 PSS schemes, when interpreted as 1.2
+			// Sig+Hash pairs, keep the hash in the
+			// signature byte.
```
FiloSottile: This is ugly as hell (but not your fault), as it's only a coincidence that the hash falls in the signature byte, the values should just be looked at as uint16. We might have to restructure how signatureAndHash works.

Lekensteyn:
Superseded by #43 and other changes that are partially upstreamed.

PR #29, opened 2017-08-08
Delegated Credentials (WORK IN PROGRESS)

tsusanka:
This is not yet finished, this PR is more of a "checkpoint".

TODO: 
- [x] saving the credentials to use them until they're valid
- [ ] test edge cases
- [ ] TLS 1.3 support
- [ ] incorporate into keyless

cc @grittygrease

### DIFF
--- a/cipher_suites.go
+++ b/cipher_suites.go
@@ -5,6 +5,7 @@
 package tls
 
 import (
+	"crypto"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/des"
@@ -33,7 +34,7 @@ type keyAgreement interface {
 
 	// This method may not be called if the server doesn't send a
 	// ServerKeyExchange message.
-	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, *x509.Certificate, *serverKeyExchangeMsg) error
+	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, crypto.PublicKey, *serverKeyExchangeMsg) error
 	generateClientKeyExchange(*Config, *clientHelloMsg, *x509.Certificate) ([]byte, *clientKeyExchangeMsg, error)
 }
 
--- a/common.go
+++ b/common.go
@@ -93,6 +93,7 @@ const (
 	extensionNextProtoNeg        uint16 = 13172 // not IANA assigned
 	extensionRenegotiationInfo   uint16 = 0xff01
 	extensionShortHeaders        uint16 = 0xff03 // Experimental
+	extensionDelegatedCredential uint16 = 0xff99 // not IANA assigned yet - https://tools.ietf.org/html/draft-rescorla-tls-subcerts-01
 )
 
 // TLS signaling cipher suite values
@@ -600,6 +601,9 @@ type Config struct {
 	// Allow short headers (experimental, and only applies to server).
 	AllowShortHeaders bool
 
+	// states if Delegated Credentials should be used
+	UseDelegatedCredentials bool
+
 	serverInitOnce sync.Once // guards calling (*Config).serverInit
 
 	// mutex protects sessionTicketKeys and originalConfig.
@@ -682,6 +686,7 @@ func (c *Config) Clone() *Config {
 		Max0RTTDataSize:             c.Max0RTTDataSize,
 		AllowShortHeaders:           c.AllowShortHeaders,
 		SessionTicketSealer:         c.SessionTicketSealer,
+		UseDelegatedCredentials:     c.UseDelegatedCredentials,
 		sessionTicketKeys:           sessionTicketKeys,
 		// originalConfig is deliberately not duplicated.
 	}
@@ -949,6 +954,9 @@ type Certificate struct {
 	// OCSPStaple contains an optional OCSP response which will be served
 	// to clients that request it.
 	OCSPStaple []byte
+	// DelegatedCredential contains a signed object containing a public
+	// key for signing handshakes
+	DelegatedCredential []byte
 	// SignedCertificateTimestamps contains an optional list of Signed
 	// Certificate Timestamps which will be served to clients that request it.
 	SignedCertificateTimestamps [][]byte
--- a/delegated_credentials.go
+++ b/delegated_credentials.go
@@ -17,7 +17,7 @@ import (
 	"time"
 )
 
-var DelegatedCredentialsIdentifier = asn1.ObjectIdentifier{2, 5, 29, 99}
+var delegatedCredentialsIdentifier = asn1.ObjectIdentifier{2, 5, 29, 99}
 
 type GetCertificate func(*ClientHelloInfo) (*Certificate, error)
 
@@ -26,13 +26,13 @@ type DelegatedCredential struct {
 	PublicKey interface{}
 }
 
-type CachedCredential struct {
+type cachedCredential struct {
 	credential []byte
 	validTime  int64
 	privateKey crypto.PrivateKey
 }
 
-type CredentialsSetup struct {
+type credentialsSetup struct {
 	scheme   SignatureScheme
 	cert     *Certificate
 	version  uint16
@@ -40,10 +40,11 @@ type CredentialsSetup struct {
 	timeFunc func() time.Time
 }
 
-var cachedCredentials map[SignatureScheme]*CachedCredential
+var cachedCredentials map[SignatureScheme]*cachedCredential
 
+// Creates new GetCertificate function which modifies the certificate
+// to include a delegated credential.
 func NewDelegatedCredentialsGetCertificate(cert *Certificate, validity time.Duration, timeFunc func() time.Time) GetCertificate {
-
 	return func(clientHelloInfo *ClientHelloInfo) (*Certificate, error) {
 		if !isCertificateValidForDelegationUsage(cert.Leaf) {
 			return nil, errors.New("tls: Delegated Credentials not supported by the certificate (DelegationUsage extension missing)")
@@ -57,7 +58,7 @@ func NewDelegatedCredentialsGetCertificate(cert *Certificate, validity time.Dura
 			return nil, errors.New("tls: Only TLS 1.2 or 1.3 are supported")
 		}
 
-		err := fetchCredentialFromCache(&CredentialsSetup{
+		err := fetchCredentialFromCache(&credentialsSetup{
 			scheme:   selectedScheme,
 			cert:     cert,
 			version:  version,
@@ -74,9 +75,11 @@ func NewDelegatedCredentialsGetCertificate(cert *Certificate, validity time.Dura
 	}
 }
 
-func fetchCredentialFromCache(setup *CredentialsSetup) (err error) {
+// Checks if a credential is already stored in memory. If not or the credential
+// is already expired a new credential is created.
+func fetchCredentialFromCache(setup *credentialsSetup) (err error) {
 	if cachedCredentials == nil {
-		cachedCredentials = make(map[SignatureScheme]*CachedCredential)
+		cachedCredentials = make(map[SignatureScheme]*cachedCredential)
 	}
 	if cachedCredentials[setup.scheme] == nil {
 		cachedCredentials[setup.scheme], err = newCredential(setup)
@@ -89,13 +92,13 @@ func fetchCredentialFromCache(setup *CredentialsSetup) (err error) {
 	return
 }
 
-func newCredential(setup *CredentialsSetup) (*CachedCredential, error) {
+func newCredential(setup *credentialsSetup) (*cachedCredential, error) {
 	credential, privateKey, err := createDelegatedCredential(setup)
 	credentialBytes, err := credential.marshalAndSign(setup.cert, setup.scheme, setup.version)
 	if err != nil {
 		return nil, fmt.Errorf("tls: creating Delegated Credential failed (%s)", err)
 	}
-	return &CachedCredential{
+	return &cachedCredential{
 		credential: credentialBytes,
 		validTime:  credential.ValidTime,
 		privateKey: privateKey,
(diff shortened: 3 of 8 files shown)


handshake_server.go
```diff
@@ -356,6 +369,24 @@ Curves:
 		}
 	}
 
+	if hs.clientHello.delegatedCredentials {
```
grittygrease: This logic should most likely be moved fully into GetCertificate. A cleaner design would be to simply create a new PrivateKey in the certificate object that corresponds to the delegated credential private key. That way almost all logic on the server side would be unchanged, aside from the addition of a new opaque delegated credentials object in the serverhello.

handshake_delegated_cred.go
```diff
@@ -0,0 +1,177 @@
...
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/asn1"
+	"errors"
+	"time"
+)
+
```
grittygrease: This tight integration with the handshake is not ideal. I think all the credential minting and verifying should be abstracted into its own package (byte slice + cert to PrivateKey and back). The only pieces that need to change on the client side are the sending of the empty extension, the extraction and validation of the delegated credential, and the CertificateVerify function.

common.go
```diff
@@ -11,6 +11,7 @@ import (
 	"crypto/rand"
 	"crypto/sha512"
 	"crypto/x509"
+	"encoding/asn1"
```
grittygrease: As mentioned below, this is not a great dependency to add. Factor out as much logic as possible into its own package and try to keep the changes to crypto/tls minimal.

common.go
```diff
@@ -949,6 +954,9 @@ type Certificate struct {
 	// OCSPStaple contains an optional OCSP response which will be served
 	// to clients that request it.
 	OCSPStaple []byte
+	// DelegatedCredential contains a signed object containing a public
+	// key for signing handshakes
+	DelegatedCredential DelegatedCredential
```
grittygrease: This is probably the wrong place for this type of structure. When the Certificate is returned from GetCertificate it should already have a serialized opaque delegated credential, not a structure that needs to be serialized.

handshake_delegated_cred_test.go
```diff
@@ -0,0 +1,283 @@
+// Copyright 2010 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
grittygrease: Again, this is too tightly integrated with the handshake. Each individual function should be unit tested with respect to its functionality, the internal tls structures should not be involved.

grittygrease:
Config.Clone and other ancillary functions are also missing the new data.

grittygrease:
Also note that the tests are failing.

common.go
```diff
@@ -614,6 +616,9 @@ type Config struct {
 	// this Config is returned by a GetConfigForClient callback. It's used
 	// by serverInit in order to copy session ticket keys if needed.
 	originalConfig *Config
+
+	// states if Delegated Credentials should be used
+	useDelegatedCredentials bool
```
tsusanka: You're right, however, how come I still can access it? https://github.com/tsusanka/tls-tris/blob/tsusanka/delegated-cred/handshake_delegated_cred_test.go#L254

common.go
```diff
@@ -959,6 +967,23 @@ type Certificate struct {
...
 }
 
+type DelegatedCredential struct {
+	ValidTime int64 // todo - change to Time?
+	PublicKey interface{}
+}
+
+// checks certificate if it contains the DelegationUsage extension
+// required for Delegated Credentials
+func (hs *serverHandshakeState) isCertificateValidForDelegationUsage() bool {
```
tsusanka: Got it. The content of the handshake_delegated_cred.go file will be then moved to separate crypto/tls/delcred (or similar name) sub package. This sub package will be responsible for creating, marshalling, signing, validating etc. of the credentials.

handshake_delegated_cred_test.go
tsusanka: Okay. So I should add unit tests - e.g. check if the credential marshalled correctly etc. What is wrong with those higher-level tests though? They still provide value, don't they?

common.go
tsusanka: I understand. The Certificate struct will then contain marshalled DC and the associated PrivateKey. Both received from the delcred sub package.

handshake_server.go
tsusanka: Thanks. I'm a little confused, how to use the GetCertificate function. If the c.GetCertificate is not null it is called to retrieve the certificate.

So, my goal is to implement a new GetCertificate function which I then "override" how the certificate is retrieved, right? The function will add the delegated credential and its PrivateKey. However, where this code should be? Also in the subpackage? Where do I then set the config to use this function?

handshake_delegated_cred.go
tsusanka: >  The only pieces that need to change on the client side are the sending of the empty extension, the extraction and validation of the delegated credential, and the CertificateVerify function.

I'm confused about the CertificateVerify part. Do you mean the CertificateVerify message? But that's used for client authentication, isn't it?

tsusanka:
Thanks for the review. I'm aware of the failing tests, it is some kind of race condition issue, I'll have a look into that once I fix the outlined issues.

handshake_server.go
tsusanka: Ah, alright, I simply set the function in the config when spinning up TLS, right?

common.go
grittygrease: Perhaps consider reusing the existing PrivateKey struct.

common.go
grittygrease: Yes, that sounds like a plan.

handshake_delegated_cred.go
grittygrease: Actually, in TLS 1.2 this is called the ServerKeyExchange.

cipher_suites.go
```diff
@@ -33,7 +34,7 @@ type keyAgreement interface {
 
 	// This method may not be called if the server doesn't send a
 	// ServerKeyExchange message.
-	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, *x509.Certificate, *serverKeyExchangeMsg) error
+	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, crypto.PublicKey, *serverKeyExchangeMsg) error
```
tsusanka: It might be possible to avoid this change by wrapping the DC's PublicKey into an x509.Certificate. However:
a) I'm not sure it'll work since we have only the PublicKey
b) all processServerKeyExchange functions use only the public key, so the dependency for x509.Certificate seems  redundant
c) the function is not exported, so it should not be a BC break (?)

tsusanka:
I think this is ready for another round of code review.
- All code is in the same _tls_ package, however, the logic is clearly separated using the GetCertificate function as discussed
- I've squashed all commits into just a few, it was a mess
- Tests are fixed
- I've implemented the support for PKCS1, it's trivial to remove it in case we decide to

cipher_suites.go
grittygrease: This is a slick change. I can't see why it wouldn't work or why *x509.Certificate was used in the first place.

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+
+type GetCertificate func(*ClientHelloInfo) (*Certificate, error)
+
+type DelegatedCredential struct {
+	ValidTime int64
+	PublicKey interface{}
+}
+
+func NewDelCredGetCertificateFunction(cert *Certificate) GetCertificate {
+
```
grittygrease: Unnecessary space?

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+var DelegatedCredentialsIdentifier = asn1.ObjectIdentifier{2, 5, 29, 99}
+
+type GetCertificate func(*ClientHelloInfo) (*Certificate, error)
+
+type DelegatedCredential struct {
+	ValidTime int64
+	PublicKey interface{}
+}
+
+func NewDelCredGetCertificateFunction(cert *Certificate) GetCertificate {
```
grittygrease: This abbreviation is a bit awkward. NewGetCertificateFromDelegatedCredential? That's also long but a bit better.

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return func(clientHelloInfo *ClientHelloInfo) (*Certificate, error) {
+
```
grittygrease: Space

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		cert.DelegatedCredential = credentialBytes
+		cert.PrivateKey = privateKey
+		return cert, nil
+	}
+}
+
+func selectVersion(versions []uint16) uint16 {
+
```
grittygrease: Newline

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+	validTill := time.Now().Add(CredentialsValidity)
+	relativeToCert := validTill.Sub(certificate.NotBefore)
+	credential = DelegatedCredential{
+		ValidTime: int64(relativeToCert.Seconds()),
+	}
+
+	if scheme == ECDSAWithP256AndSHA256 {
+		privateKey, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+		credential.PublicKey = privateKey.(crypto.Signer).Public()
+
```
grittygrease: newline

delegated_credentials.go
```diff
@@ -0,0 +1,250 @@
...
+	publicKeyBytes, err := x509.MarshalPKIXPublicKey(dc.PublicKey)
+	publicKeyLength := len(publicKeyBytes)
+	cred[4] = uint8(publicKeyLength >> 16)
+	cred[5] = uint8(publicKeyLength >> 8)
+	cred[6] = uint8(publicKeyLength)
+
+	copy(cred[7:7+publicKeyLength], publicKeyBytes)
+
+	cred[publicKeyLength+7] = uint8(scheme >> 8) // hash
```
grittygrease: This will likely be different for TLS 1.3

handshake_client.go
```diff
@@ -21,13 +21,14 @@ import (
...
-	masterSecret []byte
-	session      *ClientSessionState
+	c                *Conn
+	serverHello      *serverHelloMsg
+	hello            *clientHelloMsg
+	suite            *cipherSuite
+	finishedHash     finishedHash
+	masterSecret     []byte
+	session          *ClientSessionState
+	delCredPublicKey crypto.PublicKey
```
grittygrease: Maybe this should be just connection PublicKey, and it is set to either the delegated cred key or c.peerCertificates[0].PublicKey. Then you may be able to reduce the number of if statements below.

delegated_credentials.go
tsusanka: Yeah, I don't really like it myself. I'm not sure the `From` is accurate. So how about simply `NewDelegatedCredentialGetCertificate`?

delegated_credentials.go
```diff
@@ -0,0 +1,295 @@
...
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/asn1"
+	"errors"
+	"fmt"
+	"time"
+)
+
+// After only renewThreshold of credential's validTime is left a new one is created
+const renewThreshold = 0.2
```
tsusanka: Maybe better to reverse this? Set it to `0.8` and call it `renewAfterThreshold`

delegated_credentials.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+	return 0
+}
+
+// Creates new Delegated Credential. The type of the credential is decided based on the selected
+// SignatureScheme to ensure client's support.
+func createDelegatedCredential(config *CredentialsConfig) (credential DelegatedCredential, privateKey crypto.PrivateKey, renewAfter time.Time, err error) {
+	validTill := config.timeFunc().Add(config.Validity)
+	relativeToCert := validTill.Sub(config.Cert.Leaf.NotBefore)
```
grittygrease: The computation of the ValidTime is a rather complex dance. Is there a simpler way to compute it that is more readable?

delegated_credentials.go
```diff
@@ -0,0 +1,295 @@
...
+type CredentialsConfig struct {
+	Cert      *Certificate
+	Validity  time.Duration
+	scheme    SignatureScheme
+	version   uint16
+	timeFunc  func() time.Time
+	renewChan chan struct{}
+}
+
+var cachedCredentials map[SignatureScheme]*cachedCredential
```
grittygrease: This is a package-global variable indexed by the SignatureScheme. This means it will be shared between all instantiations, even those with different certs. You most likely want to keep this local to the CredentialsConfig.

delegated_credentials.go
```diff
@@ -0,0 +1,295 @@
...
+		cachedCredentials = make(map[SignatureScheme]*cachedCredential)
+	}
+	if cachedCredentials[config.scheme] == nil {
+		cachedCredentials[config.scheme], err = newCredential(config)
+	} else {
+		if config.timeFunc().After(cachedCredentials[config.scheme].renewAfter) {
+			go func() {
+				cachedCredentials[config.scheme], err = newCredential(config)
+				if config.renewChan != nil {
+					config.renewChan <- struct{}{}
```
grittygrease: Where is the renewChan read?

delegated_credentials.go
tsusanka: Good point! So I'm just including it as a not exported CredentialsConfig variable.

delegated_credentials.go
tsusanka: Only in [tests](https://github.com/cloudflare/tls-tris/pull/29/files#diff-74969c972e0630d3c7b131f8f82325e9R270). I was reluctant to add the channel just because of tests, but I wanted to test that the new credential is created, which I can do only after the goroutine finishes. I was talking to Salmen about this and he advised me this.

delegated_credentials.go
tsusanka: Yeah, I agree. I was thinking about it and I'm not sure there is a better way though. The code is basically the definition in itself:
> Relative time in seconds from the beginning of the certificate's notBefore value after which the Delegated Credential is no longer valid

Btw why is this relative? Why not making it a simple datetime when the credential expires? The client needs to verify the certificate's validity anyway.

----

I can remove the `relativeToCert` variable but it's not significantly better:

```go
validTill := config.timeFunc().Add(config.Validity)
renewAfter = validTill.Add(-time.Duration(config.Validity.Seconds()*renewThreshold) * time.Second)
credential = DelegatedCredential{
	ValidTime: int64(validTill.Sub(config.Cert.Leaf.NotBefore).Seconds()),
}
```

tsusanka:
Closed in favour #32

PR #34, opened 2017-09-21
tris: prevent sending 0.5RTT data

Lekensteyn:
Three patches, two of them are needed to interop as client with servers that send session tickets (#50) and are included because of contextual conflicts. The final patch is:
___
tris: prevent sending 0.5-RTT data

Disable 0.5-RTT as it has weaker security properties than 1-RTT. The
same security considerations from TLS False Start (RFC 7918) apply.

Currently the server Handshake method returns as soon as it has sent its
parameters, but it does not wait for the client to authenticate the
handshake via a Finished message. This broke a test that assumed that
the Handshake message performs a full handshake and also
(unintentionally?) enabled the server to send application data before
the handshake is complete ("0.5-RTT data").

Fix this by moving the implicit Finished message check in the handshake
message reader to the server handshake itself (previously readRecord
would process the Finished message as a side-effect of requesting
recordTypeApplicationData). And in the special case where 0-RTT data is
actually desired, process the Finished message in the ConfirmHandshake
and Read functions.

NOTE: 0.5-RTT is not disabled when the server enables 0-RTT. It is the
server responsibility to use ConfirmHandshake before writing anything.

Explicitly panic when ConfirmHandshake is used for client connections,
this is not the intended use of that API.

### DIFF
--- a/13.go
+++ b/13.go
@@ -244,11 +244,32 @@ CurvePreferenceLoop:
 	return nil
 }
 
-// readClientFinished13 is called when, on the second flight of the client,
-// a handshake message is received. This might be immediately or after the
-// early data. Once done it sends the session tickets. Under c.in lock.
-func (hs *serverHandshakeState) readClientFinished13() error {
+// readClientFinished13 is called during the server handshake (when no early
+// data it available) or after reading all early data. It discards early data if
+// the server did not accept it and then verifies the Finished message. Once
+// done it sends the session tickets. Under c.in lock.
+func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error {
 	c := hs.c
+
+	// If the client advertised and sends early data while the server does
+	// not accept it, it must be fully skipped until the Finished message.
+	for c.phase == discardingEarlyData {
+		if err := c.readRecord(recordTypeApplicationData); err != nil {
+			return err
+		}
+		// Assume receipt of Finished message (will be checked below).
+		if c.hand.Len() > 0 {
+			c.phase = waitingClientFinished
+			break
+		}
+	}
+
+	// If the client sends early data followed by a Finished message (but
+	// no end_of_early_data), the server MUST terminate the connection.
+	if c.phase != waitingClientFinished {
+		c.sendAlert(alertUnexpectedMessage)
+		return errors.New("tls: did not expect Client Finished yet")
+	}
 
 	c.phase = readingClientFinished
 	msg, err := c.readHandshake()
@@ -283,7 +304,13 @@ func (hs *serverHandshakeState) readClientFinished13() error {
 	// Any read operation after handshakeRunning and before handshakeConfirmed
 	// will be holding this lock, which we release as soon as the confirmation
 	// happens, even if the Read call might do more work.
-	c.confirmMutex.Unlock()
+	// If a Handshake is pending, c.confirmMutex will never be locked as
+	// ConfirmHandshake will wait for the handshake to complete. If a
+	// handshake was complete, and this was a confirmation, unlock
+	// c.confirmMutex now to allow readers to proceed.
+	if hasConfirmLock {
+		c.confirmMutex.Unlock()
+	}
 
 	return hs.sendSessionTicket13() // TODO: do in a goroutine
 }
--- a/conn.go
+++ b/conn.go
@@ -837,9 +837,11 @@ Again:
 
 	case recordTypeHandshake:
 		// TODO(rsc): Should at least pick off connection close.
-		if typ != want && !(c.isClient && c.config.Renegotiation != RenegotiateNever) &&
-			c.phase != waitingClientFinished {
-			return c.in.setErrorLocked(c.sendAlert(alertNoRenegotiation))
+		// If early data was being read, a Finished message is expected
+		// instead of (early) application data. Other post-handshake
+		// messages include HelloRequest and NewSessionTicket.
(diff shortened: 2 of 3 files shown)


FiloSottile:
This was very intentional. There might be protocols that will want to use 0.5RTT data, and it's not as vulnerable to replay because there is no data to replay.

Any reason to disable this protocol feature beyond fixing a test?

Lekensteyn:
While it is technically allowed by the spec, I wanted to be sure that the Handshake method follows the state machine, eventually ending up in the CONNECTED state (and authenticating the client by default).

If an application needs to send 0.5RTT data, I think it should opt-in (via an option that can be added in the future).

FiloSottile:
Define "authenticating the client", it's entirely possible I'm missing
an attack scenario. Do you mean simply making sure it's not a replayed
connection (with no data)?

Lekensteyn:
One concern is related to denial of service, an attacker can flood Client Hello and if the server believes that the client handshake has been authenticated, it might unnecessarily commit resources.

Another concern is about changing the documented, expected functionality. Handshake should run the handshake protocol and error out if things went wrong. If an attacker replays the CH, then it will only be caught during the next read.

lziest:
can we have a reference of the test we want to fix?

13.go
```diff
@@ -189,7 +210,9 @@ func (hs *serverHandshakeState) readClientFinished13() error {
 	// Any read operation after handshakeRunning and before handshakeConfirmed
 	// will be holding this lock, which we release as soon as the confirmation
 	// happens, even if the Read call might do more work.
-	c.confirmMutex.Unlock()
+	if !hasHandshakeLock {
```
lziest: call me ignorant, can you explain the logic behind this extra check? I feel this is fixing a bug. But unlock only if we have locked it, right?

Lekensteyn:
> can we have a reference of the test we want to fix?

See c03e2740 ("tris: add test for TLS 1.3 certificate validation"). The test fails because `s.Close()` closes the server socket before the client can write its Finished message.

13.go
Lekensteyn: The Read and ConfirmHandshake methods can run in parallel, that's why this lock was originally needed. If ConfirmHandshake consumes data, Read must not interfere with it.

Since the Handshake method will always be executed before the main Read and ConfirmHandshake functionality gets executed (and thus before taking the confirmMutex lock), it is safe not to acquire to lock in case readClientFinished13 is executed via the handshake (hence the new `hasHandshakeLock` name).

lziest: I will read into it, in general check then unlock always sounds racing (pun intended)

Lekensteyn:
This is essentially the same problem as TLS False Start (https://tools.ietf.org/html/rfc7918), but then in the opposite direction. Its security considerations also apply.

Lekensteyn:
New version pushed with comments added. I will probably change this when implementing draft 22 (which has contextual conflicts), so maybe review later.

Lekensteyn:
Until this is decided, I will back out the test from commit 0f32ff23edfadb5e3b9a703bc3701cfaeebe19d0 to allow client tests to pass.

grittygrease:
I'm in favor of this change. When it comes to sending data in an unauthenticated context, the conservative approach should win out unless there are extraordinary gains.

conn.go
```diff
@@ -1354,8 +1355,11 @@ func (c *Conn) ConfirmHandshake() error {
 		input = &block{data: buf.Bytes()}
 	}
 
-	for c.phase != handshakeConfirmed {
-		if err := c.readRecord(recordTypeApplicationData); err != nil {
+	// At this point, earlyDataReader has read all early data and received
+	// the end_of_early_data signal. Expect a Finished message.
+	// Locks held so far: c.confirmMutex, c.in
+	for c.phase != handshakeConfirmed { // c.phase == discardingEarlyData || c.phase == waitingClientFinished
```
grittygrease: This comment is helpful, but looks odd after the {

handshake_server.go
```diff
@@ -78,6 +78,16 @@ func (c *Conn) serverHandshake() error {
 			return err
 		}
 		c.hs = &hs
+		// If the client is sending early data while the server expects
+		// it, delay the Finished check until HandshakeConfirmed() is
+		// called or until all early data is Read(). Otherwise, complete
+		// authenticating the client now (there is no support for
+		// sending 0.50RTT data to a potential unauthenticated client).
```
grittygrease: 0.5 RTT instead of 0.50RTT?

conn.go
Lekensteyn: Fixed this and the other comment, thanks for reviewing

Lekensteyn:
Some clarifications on the last patch:
- 0.5-RTT data is **still possible** if the server enables 0-RTT. Writing will not wait for the Client Finished if 0-RTT is enabled.
- Server MUST invoke ConfirmHandshake if it wants to have replay protection (e.g. if a HTTP method is found not to be idempotent).
- As soon as the server reads more 0-RTT data than is actually available (i.e. when it does two Read calls), it WILL will wait for Client Finished. (This behavior was present before this patch.)

I also happened to came across this boringssl documentation: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Early-data
They describe that 0.5-RTT ("half-RTT") is enabled with 0-RTT which seems to match with what this PR is trying to achieve.

Lekensteyn:
(just rebased the patches on latest master, no further changes)

PR #35, opened 2017-09-21
Initial support for RSA-PSS (preparation for client)

Lekensteyn:
This is part of the changes that are required for a TLS 1.3 client.

It refactors the SAH type from TLS 1.2 to SignatureScheme in TLS 1.3, these changes have been proposed upstream:
~~https://go-review.googlesource.com/c/go/+/62210~~ (merged upstream)
~~https://go-review.googlesource.com/c/go/+/64610~~ abandoned
This PR might change based on upstream feedback, ~~but so far the only concern is about naming.~~
Based on feedback to the above abandoned patch, here is another shot at preparing for and actually implementing PSS:
https://go-review.googlesource.com/c/go/+/79735 (preparation patch)
https://go-review.googlesource.com/c/go/+/79736 (PSS support for handshake messages)
https://go-review.googlesource.com/c/go/+/79738 (RFC: enable PSS and add tests)

**EDIT**: this PR will likely be superseded by the contents of https://github.com/cloudflare/tls-tris/tree/pwu/sigalgs-pss but is pending (upstream) review!

After that, adding support for RSASSA-PSS signatures for handshake messages becomes easier.
NOTE: RSASSA-PSS is *not* advertised yet in the TLS 1.2 client handshake because it technically does not support RSASSA-PSS public keys or signatures in certificates. Furthermore, all testdata would have to be updated after such a change, so I was hesitant to do that now.

To implement full RSASSA-PSS support, upstream crypto/x509 would need to be changed and the TLS WG must settle on the required functionality (see [this list post][1]).

As a hack, a future TLS 1.3 client change will include PSS anyway in the supported_algorithms extension (but only for 1.3, not for 1.2) because the handshake will fail otherwise (PSS is currently mandatory up to at least draft -21).

 [1]: https://ietf.org/mail-archive/web/tls/current/msg24440.html

### DIFF
--- a/common.go
+++ b/common.go
@@ -174,6 +174,7 @@ type signatureType uint8
 
 const (
 	signaturePKCS1v15 signatureType = iota + 1
+	signatureRSAPSS
 	signatureECDSA
 )
 
@@ -1144,7 +1145,15 @@ func unexpectedMessageError(wanted, got interface{}) error {
 	return fmt.Errorf("tls: received unexpected handshake message of type %T when waiting for %T", got, wanted)
 }
 
+// isSupportedSignatureAlgorithm returns true if the signature algorithm is
+// allowed to be used for signing handshake messages.
 func isSupportedSignatureAlgorithm(sigAlg SignatureScheme, supportedSignatureAlgorithms []SignatureScheme) bool {
+	// TODO unhack this, PSS should be advertised in
+	// supportedSignatureAlgorithms, but that possibly needs additional
+	// support of PSS public keys and PSS signatures in certificates.
+	if sigAlg == PSSWithSHA256 || sigAlg == PSSWithSHA384 {
+		return true
+	}
 	for _, s := range supportedSignatureAlgorithms {
 		if s == sigAlg {
 			return true
@@ -1159,6 +1168,8 @@ func signatureFromSignatureScheme(signatureAlgorithm SignatureScheme) signatureT
 	switch signatureAlgorithm {
 	case PKCS1WithSHA1, PKCS1WithSHA256, PKCS1WithSHA384, PKCS1WithSHA512:
 		return signaturePKCS1v15
+	case PSSWithSHA256, PSSWithSHA384, PSSWithSHA512:
+		return signatureRSAPSS
 	case ECDSAWithSHA1, ECDSAWithP256AndSHA256, ECDSAWithP384AndSHA384, ECDSAWithP521AndSHA512:
 		return signatureECDSA
 	default:
--- a/common.go
+++ b/common.go
@@ -160,10 +160,21 @@ const (
 	// Rest of these are reserved by the TLS spec
 )
 
-// Signature algorithms for TLS 1.2 (See RFC 5246, section A.4.1)
+// Public key algorithm type.
+type pubkeyType uint8
+
 const (
-	signatureRSA   uint8 = 1
-	signatureECDSA uint8 = 3
+	keyRSA pubkeyType = iota + 1
+	keyECDSA
+)
+
+// signatureType is the generic signature algorithm, not include options like
+// the hash function.
+type signatureType uint8
+
+const (
+	signaturePKCS1v15 signatureType = iota + 1
+	signatureECDSA
 )
 
 // supportedSignatureAlgorithms contains the signature and hash algorithms that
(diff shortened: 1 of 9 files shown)


handshake_client.go
```diff
@@ -435,23 +435,26 @@ func (hs *clientHandshakeState) doFullHandshake() error {
...
 		}
 
-		var signatureType uint8
+		var signatureType signatureType
 		switch key.Public().(type) {
 		case *ecdsa.PublicKey:
 			signatureType = signatureECDSA
 		case *rsa.PublicKey:
-			signatureType = signatureRSA
+			signatureType = signatureRSAPKCS1
```
lziest: so now this value will be 0? I have a feeling something is not quite right

common.go
```diff
@@ -160,35 +160,34 @@ const (
...
+	keyRSA pubkeyType = iota
+	keyECDSA
 )
 
-// Signature algorithms for TLS 1.2 (See RFC 5246, section A.4.1)
+// A generic signature algorithm, not including any hash function.
+type signatureType uint8
+
 const (
-	signatureRSA   uint8 = 1
```
lziest: I think I missed something. Now signatureRSAPKCS1=0, RSAPSS=1, and signatureECDSA=2?

lziest:
can we push this change upstream to go master branch? I am not feeling super comfortable about remove some RFC-specific constants? I might be wrong. I need to learn.

common.go
Lekensteyn: Correct, the exact values do not exactly matter since the `signatureType` is not sent over the network, it is used only internally to track what signature verification method needs to be called.

In TLS 1.2 the SignatureAndHash (SAH) type consisted of two fields, these were merged in TLS 1.3 into a single 16-bit number. The hash type and the SAH types have been removed in the previous patch, this one removed the signature type from SAH as well.

handshake_client.go
Lekensteyn: It is not sent over the network, so the exact value does not matter. Maybe I should use `iota + 1` though to allow for "uninitialized" values to be recognized.

Lekensteyn:
Upstream had no objections for the patch, I'll poke again. Ideally the patches are reviewed+merged first in upstream to reduce the changes that we have to carry.

lziest:
Thanks, I will look again.

common.go
lziest: Make sense. I only briefly touched those code before, I didn't know the change in progress.

handshake_client.go
lziest: That could work, or we just assign 0 to some `None` constant.

lziest:
We have `signatureType`, `signatureAlgorithm`, and `signatureScheme`, those names are a little bit too confusing. I actually like the name `SigAndHash` because then I know there would be a Hash component. Now Hash are being eliminated, and Hash info are embedded in `signatureAlgorithm`? (sorry I am a little dizzy right now :-))

Lekensteyn:
I tried to follow the TLS 1.3 draft closely here, we have two types:
- signatureType: e.g. signatureECDSA
- SignatureScheme: e.g. ECDSAWithSHA256

The type is named "SignatureScheme" in TLS 1.3 because TLS 1.2 defined the "SignatureAlgorithm" type differently. But I actually use "signatureAlgorithm" as variable name since that's the draft says that throughout the document, it will refer to SignatureScheme as "SignatureAlgorithm".

It is called "SignatureScheme" because there are also signature algorithms without configurable hash, so SignatureAndHashAlgorithm would not be entirely accurate. Some combinations of hash functions and signature algorithms are also incompatible.

Lekensteyn:
By the way, I have updated the patches upstream to use iota+1 for the two types (signatureType and pubkeyType) and renamed signatureRSAPKCS1 to signaturePKCS1v15, but the functionally the patches are still the same.

lziest:
LGTM, I understand that now.

Lekensteyn:
I've rebased the patch on the latest master branch and changed the following:
- substitute the "replace signatureAndHash" and "clarify distinction" patches for three patches that do the same (including the signaturePKCS1v15 rename and iota changes mentioned before)
- renamed to "signaturePKCS1v15" in the "initial support for RSASSA-PSS" patch.
- added another fix for an issue that was discovered while running bogo client tests (namely, missing support for PSS in ServerKeyExchange)

The tree patches consist of two upstream patches (still in review) plus a change needed due to our TLS 1.3 changes (see the commit message).

Lekensteyn:
Replaced by #54

PR #37, opened 2017-09-21
Remove TLS13CipherSuites

Lekensteyn:
As suggested in #22 by @tmthrgd , the TLS13CipherSuites option was removed.

The commit messages contain more details, but the main considerations for this choice are outlined here: https://github.com/cloudflare/tls-tris/pull/22#issuecomment-329286348

Summary of change: CipherSuites can be used to configure TLS 1.3 ciphers. If CipherSuites is overridden with TLS 1.2-only ciphers, but MaxVersion is 1.3, then the default TLS 1.3 ciphers will be prepended.

### DIFF
--- a/common.go
+++ b/common.go
@@ -510,10 +510,6 @@ type Config struct {
 	// supported by the implementation.
 	CipherSuites []uint16
 
-	// TLS13CipherSuites is a list of supported cipher suites to be used in
-	// TLS 1.3. If nil, uses a list of suites supported by the implementation.
-	TLS13CipherSuites []uint16
-
 	// PreferServerCipherSuites controls whether the server selects the
 	// client's most preferred ciphersuite, or the server's most preferred
 	// ciphersuite. If true then the server's preference, as expressed in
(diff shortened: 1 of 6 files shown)


common.go
```diff
@@ -756,17 +751,30 @@ func (c *Config) time() time.Time {
...
 	s := c.CipherSuites
 	if s == nil {
 		s = defaultCipherSuites()
+	} else if c.maxVersion() >= VersionTLS13 {
+		// Ensure that TLS 1.3 suites are always present, but respect
+		// the application cipher suite preferences.
+		s13 := defaultTLS13CipherSuites()
+		if !hasOverlappingCipherSuites(s, s13) {
+			allSuites := make([]uint16, len(s13)+len(s))
+			allSuites = append(allSuites, s13...)
```
lziest: I wonder if append(allSuites, s13..., s...) will save us one line of code and make it more apparent that we push s13 ahead of s)

Lekensteyn: I tried that at first, but it is not supported: https://golang.org/ref/spec#Passing_arguments_to_..._parameters

lziest: 👍 yeah, I was going to try that myself.

Lekensteyn:
Thanks for the review, this is now in master.

PR #42, opened 2017-09-21
Disable TLS 1.3 by default, allow client to pick TLS 1.3

Lekensteyn:
Require applications to explicitly set `MaxVersion: tls.VersionTLS13` for now. Once full TLS 1.3 support is completed, the global MaxVersion can be bumped instead.

The version selection code is slightly adjusted to allow for sharing functionality between client and server.

PR #36, opened 2017-09-21
Refactor TLS 1.3 key schedule

Lekensteyn:
To enable code reuse with the TLS 1.3 client, support KeyLogWriter and prepare for supporting the label changes in draft -20, extract the current open-coded, interleaved key schedule functionality.
Hopefully the server handshake looks less complex now and easier to review for correctness.

PR #43, opened 2017-09-21
Initial TLS 1.3 client support (v2)

Lekensteyn:
Initial TLS 1.3 client support was started in #25, but there were some unsolved issues raised during review. This PR takes a different approach than #25, organizing the functionality in a different way.

Rather than completely forking the SH processing, it integrates with existing SH processing. Where possible, changes were prepared/proposed for upstream. Hopefully this makes rebasing easier.
(Unfortunately it will still conflict with golang/go@e085a891f0, I ~~can cherry-pick that change and rebase on it if desired.~~ have merged all those patches and will have to update this PR accordingly)

Other notes:
- interop test of the TLS 1.3 client with boringssl was added
- boringssl has been bumped, it's still at draft -18 but the `bssl` too used for testing has improved.
- bogo test has *not* been updated to test the client. There are too many failures that need to be taken care of (many are about unexpected alerts, like alertInternalError vs alertHandshakeFailure or alertIllegalParameter).

## Suggestion for review
Independent changes are organized in separate PRs:
- [ ] #34 - make Handshake method block until Finished is complete (needed for "tris: add test for TLS 1.3 certificate validation" to pass without race conditions) as well as ignoring NewSessionTicket to avoid handshake failure
- [ ] #35 - PSS support (required for non-ECDSA (i.e. RSA) certificates, otherwise handshake will fail)
- [x] #36 - Refactor key schedule
- [x] #37 - Remove TLS13CipherSuites (required for supporting mixed TLS 1.2/1.3 suites in CH)
- [x] #42 - Allow client to pick TLS 1.3
- [x] #49 - Fix two bogo test issues.

Since this client work is based on the other changes, those commits also duplicated into this PR. Please start reviewing this PR from "crypto/tls: allow client code to handle TLS 1.3 ServerHello.".


13.go
```diff
@@ -733,8 +733,131 @@ func (hs *serverHandshakeState) traceErr(err error) {
...
+	clientCipher, _ = hs.keySchedule.prepareCipher(secretApplicationClient)
+	serverCipher, _ = hs.keySchedule.prepareCipher(secretApplicationServer)
+	c.out.setCipher(c.vers, clientCipher)
+	c.in.setCipher(c.vers, serverCipher)
+	return nil
+}
+
+// initializeKeyShare ensures that the current keyshare matches the requested
+// group. An existing keyshare with a matching group will not be replaced.
+func (hs *clientHandshakeState) initializeKeyShare(group CurveID) error {
```
Lekensteyn: This was intended to be used for HRR, it should be removed here since it is unused.

Lekensteyn:
Superseded by #55 which adapts the code to upstream changes.

PR #45, opened 2017-10-02
Use a backing array for getSupportedVersions

tmthrgd:
This pull request eliminates the memory allocation in `(*Config).getSupportedVersions` and simplifies it substantially. It builds upon #42 by @Lekensteyn (cherry-picked as aab74cb372).

### DIFF
--- a/common.go
+++ b/common.go
@@ -836,6 +836,15 @@ func (c *Config) pickVersion(peerSupportedVersions []uint16) (uint16, bool) {
 	return 0, false
 }
 
+// configSuppVersArray is the backing array of Config.getSupportedVersions
+var configSuppVersArray = [...]uint16{VersionTLS13, VersionTLS12, VersionTLS11, VersionTLS10, VersionSSL30}
+
+// tls13DraftSuppVersArray is the backing array of Config.getSupportedVersions
+// with TLS 1.3 draft versions included.
+//
+// TODO: remove once TLS 1.3 is finalised.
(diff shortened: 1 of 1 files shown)


tmthrgd:
@Lekensteyn I've addressed the first draw back by using a separate backing array for TLS 1.3 draft versions. Adding support for any other draft versions is as simple as prepending them to the `tls13DraftSuppVersArray` array. Once the draft versions are removed this variable can be eliminated.

The second point doesn't seem like a problem to me, as it's a hypothetical. The previous code (that you wrote?) made the exact same assumption anyway as it decremented by one in all cases. (Out of interest, are you aware of any suggestion to use a non-contiguous version number?)

Lekensteyn:
Looks good, I have squashed the changes and modified the commit message (the Fixes tag hopefully helps to see which patches are fixing tris issues, and which fix upstream issues). Are you OK with this?

The second point is indeed hypothetical, I am not aware of any suggestions to use non-contiguous version numbers. For the previous version of the code, I went through some iterations but then decided to make use of the contiguous values for pragmatic reasons. If standards change, then the code can always be adjusted later 😃

tmthrgd:
@Lekensteyn That's perfectly okay with me, and I'm happy if you merge that.

Relying on contiguous leaves the final code (once the TODO's are removed) really clean imho. It's certainly easy enough to change should that ever happen. I'm glad you're on board.

Lekensteyn:
Merged, thanks for the cleanup!

Issue #52, opened 2017-11-24
Support for draft 22


Lekensteyn:
Current master supports draft 18 only. During the IETF 100 hackathon, partial support was added for "soon-to-be-draft 22" based on the pwu/client branch from #43.

While working on tests for draft22, I have found some issues:
- [x] ~~bogo: assumes draft18 support, fails for instance the `FallbackSCSV-VersionMatch-TLS13` test with "Inappropriate fallback" because our server does not recognize the advertised draft 18 version in supported_versions. Requires https://boringssl-review.googlesource.com/c/boringssl/+/23704~~ integrated into draft22 branch of shim
- [x] ~~boringssl: client requires `-tls-variant draft22`, but that uses an experimental codepoint (7e04) for  supported_version instead of 7f18 (presumably since the draft is not final yet).~~ bogo shim is now updated to use the final draft version
- [x] ~~tstclnt (NSS): fails with "tls: invalid client PSK"~~ tris fix pending, draft 21 added a ticket nonce which changed the PSK calculation
- [ ] picotls client: fails with "ptls_handshake:47" (possibly because we require X25519 while picotls seems to use P256 and sends a HRR which we do not support yet).

Branches:
- tris: pwu/draft22 - draft22 client+server developed during hackathon
- tris: pwu/server-draft22 - attempt to integrate draft22 with the server only (since client support is not merged yet)
- crypto-tls-bogo-shim: https://github.com/FiloSottile/crypto-tls-bogo-shim/tree/draft22 - update with latest bogo code that "supports" draft22 (but with above issue)

Lekensteyn:
Current master now supports client and server draft 22, minus HRR.

PR #53, opened 2017-11-24
tris: unify ServerHello processing in preparation for D22

Lekensteyn:
Merge serverHelloMsg13 into serverHelloMsg in preparation for draft 22.
This will also simplify the client implementation since only one
structure has to be checked.

Also fixed potential out-of-bounds access with keyShare unmarshal.
___
Due to the issues mentioned in #52, I will not jump straight to D22, but neither do I want to redo all client changes two times either (once to adapt for the current master, once for D22 changes).

Therefore I made this refactoring change now (it is based off "tris: update Server Hello processing for D22" from the pwu/draft22 branch, but without D22 changes). After this, I should be able to update the client patches for D18 and work in parallel on D22 changes.

### DIFF
--- a/13.go
+++ b/13.go
@@ -127,7 +127,7 @@ func (hs *serverHandshakeState) doTLS13Handshake() error {
 	config := hs.c.config
 	c := hs.c
 
-	hs.c.cipherSuite, hs.hello13.cipherSuite = hs.suite.id, hs.suite.id
+	hs.c.cipherSuite, hs.hello.cipherSuite = hs.suite.id, hs.suite.id
 	hs.c.clientHello = hs.clientHello.marshal()
 
 	// When picking the group for the handshake, priority is given to groups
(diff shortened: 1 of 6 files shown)


grittygrease:
This generally looks good. The number of if statements introduced makes me uncomfortable considering the tests are sparse. Can you make sure there are new tests that exercise these code paths?

Lekensteyn:
Most of the new branches are based off whether it is an old TLS 1.3 draft or not. If it is, then it needs to skip some fields. Otherwise, it needs to process it normally.

There is a fuzzer in handshake_messages_test.go for all messages (including this ServerHelloMsg) testing that the unmarshalling a marshalled result gives the same result. The bogo suite tends to exercise the code paths too. What other tests would you need to gain more confidence?

(Note that the current *_test.go scripts already have very little TLS 1.3 coverage.)

Lekensteyn:
Merged with approval from @Bren2010 (thanks!), I have some other changes that depend on this.

PR #56, opened 2017-11-30
tris: fix nonce length definition and actually use it

Lekensteyn:
All TLS 1.3 cipher suites so far use a nonce length of 12, but that does
not have to be the case. Correct the cipher suite definition and use it.
Spec: https://tools.ietf.org/html/draft-ietf-tls-tls13-22#section-5.3

Note: there is no functional change, the values were previously unused.

Fixes: ("[dev.tls] crypto/tls: implement TLS 1.3 cipher suites")
___
Reported in https://github.com/bifurcation/mint/issues/142


Lekensteyn:
Thanks for reviewing, merged into 759dbb355afa6c37ae9df7f9fdc864d80d45111f

PR #57, opened 2017-12-04
Draft 22 changes for the server and client

Lekensteyn:
This PR implements the changes required for draft 22 support on the server and client sides. It starts with an update to draft 21 (untested for interop) and then continues with draft 22 changes.

Dependencies:
- [x] #34 prevent sending 0.5-RTT data (three patches added due to changes related to early data handling)

### DIFF
--- a/13.go
+++ b/13.go
@@ -67,6 +67,10 @@ func newKeySchedule13(suite *cipherSuite, config *Config, clientRandom []byte) *
 func (ks *keySchedule13) setSecret(secret []byte) {
 	hash := hashForSuite(ks.suite)
 	salt := ks.secret
+	if salt != nil {
+		h0 := hash.New().Sum(nil)
+		salt = hkdfExpandLabel(hash, salt, h0, "derived", hash.Size())
+	}
 	ks.secret = hkdfExtract(hash, secret, salt)
 }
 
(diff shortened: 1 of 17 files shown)


Lekensteyn:
(just (trivially) rebased on latest master)

13.go
```diff
@@ -193,6 +197,10 @@ CurvePreferenceLoop:
 		return err
 	}
 
+	if _, err := c.writeRecord(recordTypeChangeCipherSpec, []byte{1}); err != nil {
```
grittygrease: Should this be done unilaterally? Shouldn't this be cased off of whether the request id is empty or not (signaling Middlebox Compatibility Mode)

ticket.go
```diff
@@ -171,13 +171,13 @@ func (s *sessionState13) equal(i interface{}) bool {
 		s.ageAdd == s1.ageAdd &&
 		s.createdAt == s1.createdAt &&
 		s.maxEarlyDataLen == s1.maxEarlyDataLen &&
-		bytes.Equal(s.resumptionSecret, s1.resumptionSecret) &&
+		subtle.ConstantTimeCompare(s.pskSecret, s1.pskSecret) == 1 &&
```
grittygrease: Good catch. Probably not exploitable, but worth doing in constant time.

13.go
Lekensteyn: A peer should ignore it, so other than saving a few bytes on the wire I think it is not worth it to add more logic for conditional use of CCS:
> An implementation may receive an unencrypted record of type change_cipher_spec consisting of the single byte value 0x01 at any time during the handshake and MUST simply drop it without further processing. 
> [..]
> This “compatibility mode” is not explicitly negotiated. [..] Either side can send change_cipher_spec at any time during the handshake, as they must be ignored by the peer.

Lekensteyn: I tried to make this conditionally, but bogo tests fail. Investigating, but in meantime I'll keep it unconditional.

PR #61, opened 2017-12-21
tris: remove github.com/cloudflare/go

FiloSottile:
Over time the amount of custom Go patches reduced, and tris got less
tangled to the underlying Go. Finally sever the link.

This allows more flexibility in what base Go is used (the system one),
doesn't require coordinating two repositories, and simplifies the black
magic considerably.

Make sure to use tris with Go 1.9 or higher.


FiloSottile:
CI now passing, bump.

PR #62, opened 2017-12-21
Ed25519 certificates

FiloSottile:
Ed25519 is specified in the spec https://tools.ietf.org/html/draft-ietf-curdle-pkix-07. TLS 1.3 defines the SignatureScheme for Ed25519.

This requires a patch like https://github.com/golang/go/compare/master...FiloSottile:filippo/ed25519 first, so is not a ready-to-merge PR, just some code I wanted to put out there. Feel free to close the PR or keep it around.

There's also a small bugfix commit you might want to cherry-pick.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -11,6 +11,8 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+
+	"golang_org/x/crypto/ed25519"
 )
 
 // pickSignatureAlgorithm selects a signature algorithm that is compatible with
@@ -59,6 +61,10 @@ func pickSignatureAlgorithm(pubkey crypto.PublicKey, peerSigAlgs, ourSigAlgs []S
 			if sigType == signatureECDSA {
 				return sigAlg, sigType, hashAlg, nil
 			}
+		case ed25519.PublicKey:
+			if sigType == signatureEd25519 {
+				return sigAlg, sigType, hashAlg, nil
+			}
 		}
 	}
 	return 0, 0, 0, errors.New("tls: peer doesn't support any common signature algorithms")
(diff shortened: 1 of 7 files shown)


auth.go
```diff
@@ -59,6 +61,10 @@ func pickSignatureAlgorithm(pubkey crypto.PublicKey, peerSigAlgs, ourSigAlgs []S
 			if sigType == signatureECDSA {
 				return sigAlg, sigType, hashAlg, nil
 			}
+		case ed25519.PublicKey:
+			if sigType == signatureEd25519 {
```
Lekensteyn: Note: besides TLS 1.3, this can also select Ed25519 for TLS 1.2 and is defined in:
https://tools.ietf.org/html/draft-ietf-tls-rfc4492bis

handshake_server.go
```diff
@@ -309,6 +311,11 @@ Curves:
 			hs.ecdsaOk = true
 		case *rsa.PublicKey:
 			hs.rsaSignOk = true
+		case ed25519.PublicKey:
+			if c.vers < VersionTLS13 {
+				c.sendAlert(alertInternalError)
+				return false, errors.New("tls: ed25519 only supported with version 1.3")
+			}
```
Lekensteyn: As noted before, Ed25519 is being drafted for TLS 1.2 as well, so this check is probably wrong?

I think this should set `hs.ecdsaOk = true`. It is used to check whether a ciphersuite can be selected with the configured certificate. See https://tools.ietf.org/html/draft-ietf-tls-rfc4492bis-17#page-17

>   NOTE: SignatureAlgorithm is "rsa" for the ECDHE_RSA key exchange
>   algorithm and "anonymous" for ECDH_anon.  These cases are defined in
>   TLS.  SignatureAlgorithm is "ecdsa" or "eddsa" for ECDHE_ECDSA.

FiloSottile:
This functionality is now upstream.

PR #70, opened 2018-03-01
Makefile improvement

henrydcase:
Following things where added/changed:

* Builds correctly tls-tris. Previously go had a problem with symbolic link resulting in not building tls-tris at all. I've used `rsync' instead
* Can build for selected platform with "ARCH=platform make -f Makefile"
* Possible to build from any director. Previously ``cd _dev; make `` was required, no one can ``cd /whever/you/want; make -f /tris/Makefile``


henrydcase:
Crap, not yet  ready

PR #71, opened 2018-03-01
Makefile improvement

henrydcase:
Following things where added/changed:
* Builds correctly tls-tris. Previously go had a problem with
  a symbolic link resulting in not building tls-tris at all. I've
  used `rsync' instead
* Can build for selected platform with "ARCH=platform make -f Makefile"
* Possible to build from any directory. Previously ``cd _dev; make``
  was required, it's possible to ``cd /whever/you/want; make -f /tris/Makefile``

### DIFF
--- a/_dev/Makefile
+++ b/_dev/Makefile
@@ -1,27 +1,58 @@
+# Constants
+MK_FILE_PATH = $(lastword $(MAKEFILE_LIST))
+PRJ_DIR      = $(abspath $(dir $(MK_FILE_PATH))/..)
+DEV_DIR      = $(PRJ_DIR)/_dev
+# Results will be produced in this directory (can be provided by a caller)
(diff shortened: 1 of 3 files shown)


_dev/Makefile
```diff
@@ -1,27 +1,58 @@
...
+#       I try to build here
+#GOROOT: GOROOT/$(OS_ARCH)/.ok_$(VER_OS_ARCH)
+#	rm -f GOROOT/$(OS_ARCH)/pkg/*/crypto/tls.a
+
+$(BUILD_DIR)/$(OS_ARCH)/.ok_$(VER_OS_ARCH): clean
+
+# Create clean directory structure
+	mkdir -p "$(GOROOT_LOCAL)/pkg"
+
+# Copy src/tools from system GOROOT
```
henrydcase: It would be actually better to download specific version of GO from here:
https://storage.googleapis.com/golang/go*

PR #72, opened 2018-03-02
Use go 1.10 and aligns with current state of TLS in go/crypto/tls

henrydcase:
* CI should use go 1.10 
* Align code with current state of src/crypto/tls in go repository (5158aab)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
   - docker
 
 go:
-  - 1.9.x
+  - 1.10.x
 
 env:
   - MODE=interop CLIENT=boring SERVER=boring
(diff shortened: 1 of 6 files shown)


henrydcase:
This was reviewed already by google

Lekensteyn:
Hi @henrydcase , so far I have been rebasing (initial update) / merging (later updates) from a branch based on upstream. That process is described here: https://github.com/cloudflare/tls-tris/wiki/Rebase

The intention was to make it easier to identify changes suitable for upstreaming, and identify changes that are already upstreamed.

henrydcase:
good point - I'll check it later on and align

PR #89, opened 2018-05-24
Get the code to build on Ubuntu >=18.04

cjpatton:
I needed to amend `_dev/Makefile` slightly. On my system, some of the folders in GOROOT are symbolic links to directories the Makefile tries to copy. Adding the `-H` flag to `cp` solves this, and AFAICT should not affect the build on other systems.

I also added some notes to `README.md` for getting things to run on Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,13 @@ pip install setuptools
 pip install docker
 ```
 
+* In Ubuntu (>=18.04), you'll also need to `apt install docker.io`. If docker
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -44,6 +44,13 @@ pip install setuptools
 pip install docker
 ```
 
+* In Ubuntu (>=18.04), you'll also need to `apt install docker.io`. If docker
```
henrydcase: This needs to be copy-paste'able

PR #91, opened 2018-06-02
Delegated credentials for TLS

cjpatton:
This implements two packages:
  * DC (`crypto/tls/delegated_credential`): The core functionalities of the [delegated_credentials](https://github.com/tlswg/tls-subcerts) extension for TLS. This provides an API for provisioning DCs on the backend and an API for integration into the TLS handshake on the frontend.
  * EXT (`crypto/tls/ext`): Facilitates registration of extensions for use in TLS (`crypto/tls`). Let me explain the need for this:The DC extension needs constants defined in `crypto/tls` (TLS). In turn, TLS needs to import DC, but Go disallows cyclic imports. This implements @henrydcase's suggestion to add another package for "registering" the DC extension and providing an interface for TLS. That way we have: (1) DC imports REG in order to register the DC extension. (This is done with `init()` in the DC package.); and (2)  TLS imports REG in order to invoke the DC extension.

What's left to do is integrate the DC extension into TLS. I also need to add support for PKCS1, PSS, and EcDSA signatures.


delegated_credential/delegated_credential.go
```diff
@@ -0,0 +1,167 @@
...
+	ValidTime time.Duration
+
+	// The public key of the credential.
+	PublicKey crypto.PublicKey
+
+	// The signature scheme corresponding to PublicKey.
+	scheme tls.SignatureScheme
+}
+
+// NewECDSACredential generates an ECDSA key pair for the specified group and
```
henrydcase: Comment - func is called NewCredential

delegated_credential/delegated_credential.go
```diff
@@ -0,0 +1,167 @@
...
+// UnmarshalCredential decodes a credential.
+func UnmarshalCredential(serialized []byte) (*Credential, error) {
+	// TODO(cjpatton)
+	return nil, nil
+}
+
+// Delegator stores the secret key of the delegator.
+//
+// This does not implement crypto.Signer, because this interface only works for
+// PKCS1, PSS, and ECDSA signature schemes. We also want to permit use of EdDSA,
```
henrydcase: Just a side comment: +1 for EdDSA. Last time I've tried handshake it was 25% faster than ECDSA/P256 (boringssl on arm)

henrydcase:
LGTM, good stuff!

I do prefer if unit tests for the extension are in separated folder. I would imagine E2E/integration test in a future - which probably should go somewhere else.

cjpatton:
Thanks @henrydcase, I've been working on the main chunk of the implementation on another branch. It's almost finished ... I'll merge it in a couple hours.

PR #92, opened 2018-06-04
Create package crypto/tls/constants

cjpatton:
Note: This PR is one commit ahead of #91. Review that first.

The DC extension has its own package (`crypto/tls/delegated_credential`). This is useful because its API is not only invoked by the TLS client and server, but also by the backend delegator (the entity that signs delegated credentials for use by the TLS server). Thus, if we wanted to implement DC in the `crypto/tls` package, we would need to change its exported API. It seems cleaner to leave this API alone.
 
The problem is that the DC extension needs to import some constants from `crypto/tls`: in particular the identifiers for signature schemes and the protocol version. But since `crypto/tls` imports
`crypto/tls/delegated_credential`, there is a circular dependency. This PR removes this circular dependency by defining these constants in a new package `crypto/tls/constants`.

This solution is somewhat inelegant in that we still need the names of the constants in the `tls` package. In fact, there are external packages in golang that use names like `tls.VersionTLS12`.


common.go
```diff
@@ -22,14 +23,14 @@ import (
 )
 
 const (
-	VersionSSL30        = 0x0300
```
henrydcase: One of the main points is to keep this code as close as master branch in go lang, so that it's easy to upstream (that's we keep TLS 1.3 in separated file).
I understand where you want to go, but big refactoring is probably not what we need at this point.

henrydcase:
I don't think introducing new package _in TLS lib_ in order to break dependency cycle isn't right thing to do. 

In C you would simply do " #ifndef #define #endif" to solve this particular problem. In go, "interface" is a first thing which comes to my mind. Something like: 

* package DC
** contains all implementation 
** imports "tls"
** implements package initialization ("func init()"). Package calls "dc_interface.RegisterExtension(Extension_ID, NewDCExtension())

* package DC_Interface
** doesn't import neither tls nor dc
** implements DCextension interface
** Implements "RegisterExtension()"
** Implements "Extension()" -> return type is an interface. It returns object which implements the interface.

* package "tls"
** imports DC_Interface

This way you break dependency and don't need to change "tls". It's also DIP from SOLID.
There is maybe one more thing which is bottering me, but I need to read DC RFC first.

common.go
```diff
@@ -22,14 +23,14 @@ import (
...
-	VersionTLS10        = 0x0301
-	VersionTLS11        = 0x0302
-	VersionTLS12        = 0x0303
-	VersionTLS13        = 0x0304
-	VersionTLS13Draft18 = 0x7f00 | 18
-	VersionTLS13Draft21 = 0x7f00 | 21
-	VersionTLS13Draft22 = 0x7f00 | 22
+	VersionSSL30        = constants.VersionSSL30
```
henrydcase: The correct number of definitions for "VersionSSL30" variable is 1 or less :) 
("or less" being prefered:)

cjpatton:
Thanks for the comments, @henrydcase! I'm gonna dump this PR and start anew. I'll ping you when it's ready.

PR #95, opened 2018-06-15
Delegated credentials for TLS

cjpatton:
This implements the [delegated_credential](https://tools.ietf.org/html/draft-ietf-tls-subcerts) extension for TLS 1.2 and 1.3.

### DIFF
--- a/13.go
+++ b/13.go
@@ -403,6 +403,16 @@ func (hs *serverHandshakeState) sendCertificate13() error {
 	if len(certEntries) > 0 && hs.clientHello.scts {
 		certEntries[0].sctList = hs.cert.SignedCertificateTimestamps
 	}
+
+	// If hs.delegatedCredential is set (see hs.readClientHello()) then the
+	// server is using the delegated credential extension. In TLS 1.3, the DC is
+	// added as an extension to the end-entity certificate, i.e., the last
+	// CertificateEntry of Certificate.certficate_list (see
+	// https://tools.ietf.org/html/draft-ietf-tls-subcerts).
+	if len(certEntries) > 0 && hs.clientHello.delegatedCredential && hs.delegatedCredential != nil {
+		certEntries[0].delegatedCredential = hs.delegatedCredential
+	}
+
 	certMsg := &certificateMsg13{certificates: certEntries}
 
 	hs.keySchedule.write(certMsg.marshal())
@@ -423,7 +433,7 @@ func (hs *serverHandshakeState) sendCertificate13() error {
 	}
 
 	toSign := prepareDigitallySigned(sigHash, "TLS 1.3, server CertificateVerify", hs.keySchedule.transcriptHash.Sum(nil))
-	signature, err := hs.cert.PrivateKey.(crypto.Signer).Sign(c.config.rand(), toSign[:], opts)
+	signature, err := hs.privateKey.(crypto.Signer).Sign(c.config.rand(), toSign[:], opts)
 	if err != nil {
 		c.sendAlert(alertInternalError)
 		return err
@@ -468,9 +478,9 @@ func (c *Conn) handleEndOfEarlyData() error {
 // See https://tools.ietf.org/html/draft-ietf-tls-tls13-18#section-4.4.1.2
 func (hs *serverHandshakeState) selectTLS13SignatureScheme() (sigScheme SignatureScheme, err error) {
 	var supportedSchemes []SignatureScheme
-	signer, ok := hs.cert.PrivateKey.(crypto.Signer)
+	signer, ok := hs.privateKey.(crypto.Signer)
 	if !ok {
-		return 0, errors.New("tls: certificate private key does not implement crypto.Signer")
+		return 0, errors.New("tls: private key does not implement crypto.Signer")
 	}
 	pk := signer.Public()
 	if _, ok := pk.(*rsa.PublicKey); ok {
@@ -1034,10 +1044,32 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
 		return unexpectedMessageError(certMsg, msg)
 	}
 	hs.keySchedule.write(certMsg.marshal())
+
 	// Validate certificates.
 	certs := getCertsFromEntries(certMsg.certificates)
 	if err := hs.processCertsFromServer(certs); err != nil {
 		return err
+	}
+
+	// Validate the DC if present. The DC is only processed if the extension was
+	// indicated by the ClientHello; otherwise this call will result in an
+	// "illegal_parameter" alert. The call also asserts that the DC extension
+	// did not appear in the ServerHello.
+	if len(certMsg.certificates) > 0 {
+		if err := hs.processDelegatedCredentialFromServer(
+			certMsg.certificates[0].delegatedCredential); err != nil {
+			return err
+		}
+	}
+
+	// Set the public key used to verify the handshake.
+	pk := hs.c.peerCertificates[0].PublicKey
+
+	// If the delegated credential extension has successfully been negotiated,
+	// then the  CertificateVerify signature will have been produced with the
+	// DelegatedCredential's private key.
+	if hs.c.verifiedDc != nil {
+		pk = hs.c.verifiedDc.PublicKey
 	}
 
 	// Receive CertificateVerify message.
@@ -1052,7 +1084,7 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
 	}
 	err, alertCode := verifyPeerHandshakeSignature(
 		certVerifyMsg,
-		hs.c.peerCertificates[0].PublicKey,
+		pk,
 		hs.hello.supportedSignatureAlgorithms,
 		hs.keySchedule.transcriptHash.Sum(nil),
 		"TLS 1.3, server CertificateVerify")
--- a/cipher_suites.go
+++ b/cipher_suites.go
@@ -5,6 +5,7 @@
 package tls
 
 import (
+	"crypto"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/des"
@@ -12,7 +13,6 @@ import (
 	"crypto/rc4"
 	"crypto/sha1"
 	"crypto/sha256"
-	"crypto/x509"
 	"hash"
 
 	"golang_org/x/crypto/chacha20poly1305"
@@ -26,15 +26,15 @@ type keyAgreement interface {
 	// In the case that the key agreement protocol doesn't use a
 	// ServerKeyExchange message, generateServerKeyExchange can return nil,
 	// nil.
-	generateServerKeyExchange(*Config, *Certificate, *clientHelloMsg, *serverHelloMsg) (*serverKeyExchangeMsg, error)
-	processClientKeyExchange(*Config, *Certificate, *clientKeyExchangeMsg, uint16) ([]byte, error)
+	generateServerKeyExchange(*Config, crypto.PrivateKey, *clientHelloMsg, *serverHelloMsg) (*serverKeyExchangeMsg, error)
+	processClientKeyExchange(*Config, crypto.PrivateKey, *clientKeyExchangeMsg, uint16) ([]byte, error)
 
 	// On the client side, the next two methods are called in order.
 
 	// This method may not be called if the server doesn't send a
 	// ServerKeyExchange message.
-	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, *x509.Certificate, *serverKeyExchangeMsg) error
-	generateClientKeyExchange(*Config, *clientHelloMsg, *x509.Certificate) ([]byte, *clientKeyExchangeMsg, error)
+	processServerKeyExchange(*Config, *clientHelloMsg, *serverHelloMsg, crypto.PublicKey, *serverKeyExchangeMsg) error
+	generateClientKeyExchange(*Config, *clientHelloMsg, crypto.PublicKey) ([]byte, *clientKeyExchangeMsg, error)
 }
 
 const (
--- a/common.go
+++ b/common.go
@@ -97,6 +97,7 @@ const (
 	extensionKeyShare                uint16 = 51
 	extensionNextProtoNeg            uint16 = 13172 // not IANA assigned
 	extensionRenegotiationInfo       uint16 = 0xff01
+	extensionDelegatedCredential     uint16 = 0xff02 // TODO(any) Get IANA assignment
 )
 
 // TLS signaling cipher suite values
@@ -356,6 +357,10 @@ type ClientHelloInfo struct {
 	// immediately available for Read.
 	Offered0RTTData bool
 
+	// AcceptsDelegatedCredential is true if the client indicated willingness
+	// to negotiate the delegated credential extension.
+	AcceptsDelegatedCredential bool
+
 	// The Fingerprint is an sequence of bytes unique to this Client Hello.
 	// It can be used to prevent or mitigate 0-RTT data replays as it's
 	// guaranteed that a replayed connection will have the same Fingerprint.
@@ -609,6 +614,22 @@ type Config struct {
 	// session tickets, instead of SessionTicketKey.
 	SessionTicketSealer SessionTicketSealer
 
+	// AcceptDelegatedCredential is true if the client is willing to negotiate
+	// the delegated credential extension.
+	//
+	// This value has no meaning for the server.
+	//
+	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
+	AcceptDelegatedCredential bool
+
+	// GetDelegatedCredential returns a DelegatedCredential for use with the
+	// delegated credential extension based on the ClientHello and TLS version
+	// selected for the session. If this is nil, then the server will not offer
+	// a DelegatedCredential.
+	//
+	// This value has no meaning for the client.
+	GetDelegatedCredential func(*ClientHelloInfo, uint16) (*DelegatedCredential, crypto.PrivateKey, error)
+
 	serverInitOnce sync.Once // guards calling (*Config).serverInit
 
 	// mutex protects sessionTicketKeys.
@@ -685,6 +706,8 @@ func (c *Config) Clone() *Config {
 		Accept0RTTData:              c.Accept0RTTData,
 		Max0RTTDataSize:             c.Max0RTTDataSize,
 		SessionTicketSealer:         c.SessionTicketSealer,
+		AcceptDelegatedCredential:   c.AcceptDelegatedCredential,
+		GetDelegatedCredential:      c.GetDelegatedCredential,
 		sessionTicketKeys:           sessionTicketKeys,
 	}
 }
--- a/conn.go
+++ b/conn.go
@@ -51,11 +51,14 @@ type Conn struct {
 	didResume        bool // whether this connection was a session resumption
 	cipherSuite      uint16
 	ocspResponse     []byte   // stapled OCSP response
-	scts             [][]byte // signed certificate timestamps from server
+	scts             [][]byte // Signed certificate timestamps from server
 	peerCertificates []*x509.Certificate
 	// verifiedChains contains the certificate chains that we built, as
 	// opposed to the ones presented by the server.
 	verifiedChains [][]*x509.Certificate
+	// verifiedDc is set by a client who negotiates the use of a valid delegated
+	// credential.
+	verifiedDc *DelegatedCredential
 	// serverName contains the server name indicated by the client, if any.
 	serverName string
 	// secureRenegotiation is true if the server echoed the secure
--- a/generate_cert.go
+++ b/generate_cert.go
@@ -14,6 +14,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
@@ -34,6 +35,7 @@ var (
 	isCA       = flag.Bool("ca", false, "whether this cert should be its own Certificate Authority")
 	rsaBits    = flag.Int("rsa-bits", 2048, "Size of RSA key to generate. Ignored if --ecdsa-curve is set")
 	ecdsaCurve = flag.String("ecdsa-curve", "", "ECDSA curve to use to generate a key. Valid values are P224, P256 (recommended), P384, P521")
+	isDC       = flag.Bool("dc", false, "whether this cert can be used with delegated credentials")
 )
 
 func publicKey(priv interface{}) interface{} {
@@ -135,6 +137,11 @@ func main() {
 	if *isCA {
 		template.IsCA = true
 		template.KeyUsage |= x509.KeyUsageCertSign
+	}
+
+	if *isDC {
+		template.ExtraExtensions = append(
+			template.ExtraExtensions, *tls.CreateDelegationUsagePKIXExtension())
 	}
 
 	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, publicKey(priv), priv)
--- a/handshake_client.go
+++ b/handshake_client.go
@@ -65,6 +65,7 @@ func makeClientHello(config *Config) (*clientHelloMsg, error) {
 		supportedPoints:              []uint8{pointFormatUncompressed},
 		nextProtoNeg:                 len(config.NextProtos) > 0,
 		secureRenegotiationSupported: true,
+		delegatedCredential:          config.AcceptDelegatedCredential,
 		alpnProtocols:                config.NextProtos,
 	}
 	possibleCipherSuites := config.cipherSuites()
@@ -412,6 +413,50 @@ func (hs *clientHandshakeState) processCertsFromServer(certificates [][]byte) er
 	return nil
 }
 
+// processDelegatedCredentialFromServer unmarshals the DelegatedCredential
+// offered by the server (if present) and validates it using the peer
+// certificate.
+func (hs *clientHandshakeState) processDelegatedCredentialFromServer(dc []byte) error {
+	c := hs.c
+
+	var cred *DelegatedCredential
+	var err error
+	if dc != nil {
+
+		// Assert that the DC extension was indicated by the client.
+		if !hs.hello.delegatedCredential {
+			c.sendAlert(alertUnexpectedMessage)
+			return errors.New("tls: got delegated credential extension without indication")
+		}
+
+		// Assert that the DC was sent in the ServerHello in (and only in)
+		// version 1.2.
+		if hs.serverHello.delegatedCredential != nil && hs.serverHello.vers != VersionTLS12 {
+			c.sendAlert(alertIllegalParameter)
+			return errors.New("tls: ServerHello with delegated credential extension in TLS != 1.2")
+		}
+
+		cred, err = UnmarshalDelegatedCredential(dc)
+		if err != nil {
+			c.sendAlert(alertDecodeError)
+			return fmt.Errorf("tls: delegated credential: %s", err)
+		}
+	}
+
+	if cred != nil && !c.config.InsecureSkipVerify {
+		if v, err := cred.Validate(c.peerCertificates[0], hs.c.vers, c.config.time()); err != nil {
+			c.sendAlert(alertIllegalParameter)
+			return fmt.Errorf("delegated credential: %s", err)
+		} else if !v {
+			c.sendAlert(alertIllegalParameter)
+			return errors.New("delegated credential: signature invalid")
+		}
+	}
+
+	c.verifiedDc = cred
+	return nil
+}
+
 func (hs *clientHandshakeState) doFullHandshake() error {
 	c := hs.c
 
@@ -430,6 +475,13 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		// If this is the first handshake on a connection, process and
 		// (optionally) verify the server's certificates.
 		if err := hs.processCertsFromServer(certMsg.certificates); err != nil {
+			return err
+		}
+		// Validate the DC if present. The DC is only processed if the
+		// extension was indicated by the ClientHello; otherwise this call will
+		// result in an "illegal_parameter" alert. It also asserts that the DC
+		// was sent in the ServerHello if and only if TLS 1.2 is in use.
+		if err := hs.processDelegatedCredentialFromServer(hs.serverHello.delegatedCredential); err != nil {
 			return err
 		}
 	} else {
@@ -477,10 +529,20 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 
 	keyAgreement := hs.suite.ka(c.vers)
 
+	// Set the public key used to verify the handshake.
+	pk := c.peerCertificates[0].PublicKey
+
+	// If the delegated credential extension has successfully been negotiated,
+	// then the ServerKeyExchange DelegatedCredential's private key.
+	if c.verifiedDc != nil {
+		pk = c.verifiedDc.PublicKey
+	}
+
 	skx, ok := msg.(*serverKeyExchangeMsg)
 	if ok {
 		hs.finishedHash.Write(skx.marshal())
-		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, c.peerCertificates[0].PublicKey, skx)
+
+		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, pk, skx)
 		if err != nil {
 			c.sendAlert(alertUnexpectedMessage)
 			return err
@@ -529,7 +591,7 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		}
 	}
 
-	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(c.config, hs.hello, c.peerCertificates[0].PublicKey)
+	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(c.config, hs.hello, pk)
 	if err != nil {
 		c.sendAlert(alertInternalError)
 		return err
--- a/handshake_client.go
+++ b/handshake_client.go
@@ -480,7 +480,7 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 	skx, ok := msg.(*serverKeyExchangeMsg)
 	if ok {
 		hs.finishedHash.Write(skx.marshal())
-		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, c.peerCertificates[0], skx)
+		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, c.peerCertificates[0].PublicKey, skx)
 		if err != nil {
 			c.sendAlert(alertUnexpectedMessage)
 			return err
@@ -529,7 +529,7 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		}
 	}
 
-	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(c.config, hs.hello, c.peerCertificates[0])
+	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(c.config, hs.hello, c.peerCertificates[0].PublicKey)
 	if err != nil {
 		c.sendAlert(alertInternalError)
 		return err
--- a/handshake_messages.go
+++ b/handshake_messages.go
@@ -49,12 +49,14 @@ type clientHelloMsg struct {
 	psks                             []psk
 	pskKeyExchangeModes              []uint8
 	earlyData                        bool
+	delegatedCredential              bool
 }
 
-// Helpers
-// Function used for signature_algorithms and signature_algorithrms_cert extensions only
-// (for more details, see TLS 1.3 draft 28, 4.2.3)
-// Function advances data slice and returns it, so that it can be used for further processing
+// Function used for signature_algorithms and signature_algorithrms_cert
+// extensions only (for more details, see TLS 1.3 draft 28, 4.2.3).
+//
+// It advances data slice and returns it, so that it can be used for further
+// processing
 func marshalExtensionSignatureAlgorithms(extension uint16, data []byte, schemes []SignatureScheme) []byte {
 	algNum := uint16(len(schemes))
 	if algNum == 0 {
@@ -126,7 +128,8 @@ func (m *clientHelloMsg) equal(i interface{}) bool {
 		eqStrings(m.alpnProtocols, m1.alpnProtocols) &&
 		eqKeyShares(m.keyShares, m1.keyShares) &&
 		eqUint16s(m.supportedVersions, m1.supportedVersions) &&
-		m.earlyData == m1.earlyData
+		m.earlyData == m1.earlyData &&
+		m.delegatedCredential == m1.delegatedCredential
 }
 
 func (m *clientHelloMsg) marshal() []byte {
@@ -200,6 +203,9 @@ func (m *clientHelloMsg) marshal() []byte {
 		numExtensions++
 	}
 	if m.earlyData {
+		numExtensions++
+	}
+	if m.delegatedCredential {
 		numExtensions++
 	}
 	if numExtensions > 0 {
@@ -419,6 +425,10 @@ func (m *clientHelloMsg) marshal() []byte {
 		z[1] = byte(extensionEarlyData)
 		z = z[4:]
 	}
+	if m.delegatedCredential {
+		binary.BigEndian.PutUint16(z, extensionDelegatedCredential)
+		z = z[4:]
+	}
 
 	m.raw = x
 
@@ -483,6 +493,7 @@ func (m *clientHelloMsg) unmarshal(data []byte) alert {
 	m.psks = nil
 	m.pskKeyExchangeModes = nil
 	m.earlyData = false
+	m.delegatedCredential = false
 
 	if len(data) == 0 {
 		// ClientHello is optionally followed by extension data
@@ -747,6 +758,9 @@ func (m *clientHelloMsg) unmarshal(data []byte) alert {
 		case extensionEarlyData:
 			// https://tools.ietf.org/html/draft-ietf-tls-tls13-18#section-4.2.8
 			m.earlyData = true
+		case extensionDelegatedCredential:
+			// https://tools.ietf.org/html/draft-ietf-tls-subcerts
+			m.delegatedCredential = true
 		}
 		data = data[length:]
 		bindersOffset += length
@@ -774,6 +788,10 @@ type serverHelloMsg struct {
 	secureRenegotiation          []byte
 	secureRenegotiationSupported bool
 	alpnProtocol                 string
+
+	// TLS 1.2. In TLS 1.3, the DC extension is included in of the end-entity
+	// certificate in the Certificate message.
+	delegatedCredential []byte
 
 	// TLS 1.3
 	keyShare    keyShare
@@ -810,6 +828,7 @@ func (m *serverHelloMsg) equal(i interface{}) bool {
 		bytes.Equal(m.secureRenegotiation, m1.secureRenegotiation) &&
 		m.alpnProtocol == m1.alpnProtocol &&
 		m.keyShare.group == m1.keyShare.group &&
+		bytes.Equal(m.delegatedCredential, m1.delegatedCredential) &&
 		bytes.Equal(m.keyShare.data, m1.keyShare.data) &&
 		m.psk == m1.psk &&
 		m.pskIdentity == m1.pskIdentity
@@ -861,6 +880,10 @@ func (m *serverHelloMsg) marshal() []byte {
 			sctLen += len(sct) + 2
 		}
 		extensionsLength += 2 + sctLen
+		numExtensions++
+	}
+	if dcLen := len(m.delegatedCredential); dcLen > 0 && m.vers == VersionTLS12 {
+		extensionsLength += 4 + dcLen
 		numExtensions++
 	}
 	if m.keyShare.group != 0 {
@@ -992,7 +1015,13 @@ func (m *serverHelloMsg) marshal() []byte {
 			z = z[len(sct)+2:]
 		}
 	}
-
+	if dcLen := len(m.delegatedCredential); dcLen > 0 && m.vers == VersionTLS12 {
+		binary.BigEndian.PutUint16(z, extensionDelegatedCredential)
+		binary.BigEndian.PutUint16(z[2:], uint16(dcLen))
+		z = z[4:]
+		copy(z, m.delegatedCredential)
+		z = z[dcLen:]
+	}
 	if m.keyShare.group != 0 {
 		z[0] = uint8(extensionKeyShare >> 8)
 		z[1] = uint8(extensionKeyShare)
@@ -1180,6 +1209,11 @@ func (m *serverHelloMsg) unmarshal(data []byte) alert {
 				m.scts = append(m.scts, d[:sctLen])
 				d = d[sctLen:]
 			}
+		case extensionDelegatedCredential:
+			if m.vers != VersionTLS12 {
+				return alertUnexpectedMessage
+			}
+			m.delegatedCredential = data[:length]
 		case extensionKeyShare:
 			d := data[:length]
 
@@ -1422,9 +1456,10 @@ func (m *certificateMsg) unmarshal(data []byte) alert {
 }
 
 type certificateEntry struct {
-	data       []byte
-	ocspStaple []byte
-	sctList    [][]byte
+	data                []byte
+	ocspStaple          []byte
+	sctList             [][]byte
+	delegatedCredential []byte
 }
 
 type certificateMsg13 struct {
@@ -1446,6 +1481,7 @@ func (m *certificateMsg13) equal(i interface{}) bool {
 		ok := bytes.Equal(m.certificates[i].data, m1.certificates[i].data)
 		ok = ok && bytes.Equal(m.certificates[i].ocspStaple, m1.certificates[i].ocspStaple)
 		ok = ok && eqByteSlices(m.certificates[i].sctList, m1.certificates[i].sctList)
+		ok = ok && bytes.Equal(m.certificates[i].delegatedCredential, m1.certificates[i].delegatedCredential)
 		if !ok {
 			return false
 		}
@@ -1471,6 +1507,9 @@ func (m *certificateMsg13) marshal() (x []byte) {
 			for _, sct := range cert.sctList {
 				i += 2 + len(sct)
 			}
+		}
+		if len(cert.delegatedCredential) != 0 {
+			i += 4 + len(cert.delegatedCredential)
 		}
 	}
 
@@ -1546,6 +1585,15 @@ func (m *certificateMsg13) marshal() (x []byte) {
 			sctLenPos[2] = uint8(sctLen >> 8)
 			sctLenPos[3] = uint8(sctLen)
 		}
+		if len(cert.delegatedCredential) != 0 {
+			binary.BigEndian.PutUint16(z, extensionDelegatedCredential)
+			binary.BigEndian.PutUint16(z[2:], uint16(len(cert.delegatedCredential)))
+			z = z[4:]
+			copy(z, cert.delegatedCredential)
+			z = z[len(cert.delegatedCredential):]
+			extensionLen += 4 + len(cert.delegatedCredential)
+		}
+
 		extLenPos[0] = uint8(extensionLen >> 8)
 		extLenPos[1] = uint8(extensionLen)
 	}
@@ -1651,6 +1699,8 @@ func (m *certificateMsg13) unmarshal(data []byte) alert {
 					m.certificates[i].sctList = append(m.certificates[i].sctList, body[2:2+sctLen])
 					body = body[2+sctLen:]
 				}
+			case extensionDelegatedCredential:
+				m.certificates[i].delegatedCredential = body
 			}
 		}
 	}
--- a/handshake_server.go
+++ b/handshake_server.go
@@ -479,7 +479,7 @@ func (hs *serverHandshakeState) doFullHandshake() error {
 	}
 
 	keyAgreement := hs.suite.ka(c.vers)
-	skx, err := keyAgreement.generateServerKeyExchange(c.config, hs.cert, hs.clientHello, hs.hello)
+	skx, err := keyAgreement.generateServerKeyExchange(c.config, hs.cert.PrivateKey, hs.clientHello, hs.hello)
 	if err != nil {
 		c.sendAlert(alertHandshakeFailure)
 		return err
@@ -572,7 +572,7 @@ func (hs *serverHandshakeState) doFullHandshake() error {
 	}
 	hs.finishedHash.Write(ckx.marshal())
 
-	preMasterSecret, err := keyAgreement.processClientKeyExchange(c.config, hs.cert, ckx, c.vers)
+	preMasterSecret, err := keyAgreement.processClientKeyExchange(c.config, hs.cert.PrivateKey, ckx, c.vers)
 	if err != nil {
 		if err == errClientKeyExchange {
 			c.sendAlert(alertDecodeError)
--- a/handshake_server.go
+++ b/handshake_server.go
@@ -30,6 +30,11 @@ type serverHandshakeState struct {
 	clientHello           *clientHelloMsg
 	hello                 *serverHelloMsg
 	cert                  *Certificate
+	privateKey            crypto.PrivateKey
+
+	// A marshalled DelegatedCredential to be sent to the client in the
+	// handshake.
+	delegatedCredential []byte
 
 	// TLS 1.0-1.2 fields
 	ellipticOk      bool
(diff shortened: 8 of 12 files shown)


13.go
```diff
@@ -402,10 +402,20 @@ func (hs *serverHandshakeState) sendCertificate13() error {
...
+	var err error
+	if len(certEntries) > 0 && hs.clientHello.delegatedCredential && hs.dc != nil {
+		certEntries[0].dc = hs.dc
+	}
+
 	certMsg := &certificateMsg13{certificates: certEntries}
 
 	hs.keySchedule.write(certMsg.marshal())
-	if _, err := c.writeRecord(recordTypeHandshake, certMsg.marshal()); err != nil {
+	if _, err = c.writeRecord(recordTypeHandshake, certMsg.marshal()); err != nil {
```
Bren2010: no

common.go
```diff
@@ -969,18 +992,22 @@ var writerMutex sync.Mutex
 // A Certificate is a chain of one or more certificates, leaf first.
 type Certificate struct {
 	Certificate [][]byte
+
```
Bren2010: We should keep the diff as conservative as possible, so I would remove these whitespace changes along with others in other files

conn.go
```diff
@@ -47,17 +47,25 @@ type Conn struct {
...
+	ocspResponse []byte   // stapled OCSP response
+	scts         [][]byte // Signed certificate timestamps from server
+
 	peerCertificates []*x509.Certificate
+
 	// verifiedChains contains the certificate chains that we built, as
 	// opposed to the ones presented by the server.
 	verifiedChains [][]*x509.Certificate
+
+	// validatedDs is set by a client who negotiates the use of a valid
```
Bren2010: s -> validatedDc

generate_cert.go
```diff
@@ -137,6 +139,11 @@ func main() {
 		template.KeyUsage |= x509.KeyUsageCertSign
 	}
 
+	if *isDC {
+		template.ExtraExtensions = append(
+			template.ExtraExtensions, *tls.CreateDelegationUsagePKIXExtension())
```
Bren2010: Maybe instead of this, which is a new pattern, you add a field to x509.Certificate called PermitsDelegatedCredentials and rig up the machinery to create the extension for you.

common.go
```diff
@@ -607,6 +612,22 @@ type Config struct {
...
+	//
+	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
+	AcceptDelegatedCredential bool
+
+	// GetDelegatedCredential returns a DelegatedCredential for use with the
+	// delegated_credential extension based on the ClientHello and TLS version
+	// selected for the session. If this is nil, then the server will not offer
+	// a DelegatedCredential.
+	//
+	// This value has no meaning for the client.
```
Bren2010: Hm, are delegated creds for client certs a thing?

handshake_client.go
```diff
@@ -624,6 +686,14 @@ func (hs *clientHandshakeState) serverResumedSession() bool {
 func (hs *clientHandshakeState) processServerHello() (bool, error) {
 	c := hs.c
 
+	// Process the delegated_credential extension.
+	if hs.serverHello.dc != nil {
+		if hs.serverHello.vers != VersionTLS12 {
+			c.sendAlert(alertIllegalParameter)
+			return false, errors.New("tls: ServerHello with delegated_credential extension in TLS != 1.2")
+		}
```
Bren2010: Is there a reason you use an underscore in delegated_credential? (ie, did you see that pattern somewhere?) It strikes me as odd

handshake_messages.go
```diff
@@ -34,31 +34,32 @@ type clientHelloMsg struct {
...
+	l -= 2
+	data[0] = byte(l >> 8)
+	data[1] = byte(l)
+	data = data[2:]
+	for _, sigAlgo := range sigSchemes {
+		data[0] = byte(sigAlgo >> 8)
+		data[1] = byte(sigAlgo)
+		data = data[2:]
+	}
+	return data
```
Bren2010: 😭

Bren2010: also, marshal is spelt with one L

handshake_messages.go
```diff
@@ -407,6 +412,11 @@ func (m *clientHelloMsg) marshal() []byte {
 		z[1] = byte(extensionEarlyData)
 		z = z[4:]
 	}
+	if m.delegatedCredential {
+		z[0] = byte(extensionDelegatedCredential >> 8)
+		z[1] = byte(extensionDelegatedCredential & 0xff)
```
Bren2010: the two times when this is done before, there's no & 0xff

common.go
cjpatton: They're not in the spec, no.

generate_cert.go
cjpatton: crypto/x509 is not part of tris, just crypto/tls

handshake_client.go
cjpatton: In the context of the TLS standard, the extension is identified by the string "delegated_credential".

handshake_messages.go
cjpatton: Yes, because the other extension numbers are less than 256 and so fit in a single byte; ours doesn't. For example, `byte(0xffff)` will give you an error, but `byte(0xff)` will not.

handshake_messages.go
cjpatton: What exactly is the complaint here? This tab was added by vim's built-in linter, so I assume this change is "idomatic". I'd have to disable my linter to fix this.

Thx for noticing the spelling.

13.go
cjpatton: TODO

common.go
cjpatton: Ack.

conn.go
cjpatton: Ack.

handshake_messages.go
Bren2010: I'm glad you fixed it, sad it happened.

handshake_messages.go
Bren2010: https://play.golang.org/p/N5GsqbuKr91

13.go
cjpatton: Done.

common.go
cjpatton: Done.

conn.go
cjpatton: Done.

handshake_messages.go
cjpatton: I did `s/arshall/arshal/g` in `handshake_messages.go`. Note that this spelling error existed before I touched the code. So technically this change makes the PR not minimal

cjpatton:
Ready for round two, @Bren2010.

handshake_messages.go
cjpatton: Hmmm ... I upgraded to 1.10.3 and got the same error. Have you tried running this on your machine?

Bren2010: I see... it's doing compile-time overflow checking because it's a const, not a var. And the others don't need &0xff because they're genuinely less than 2^8. I'm sorry, disregard this

cjpatton: Yeah, I just figured out the same :)

https://play.golang.org/p/jpFT2CDPV-i

cjpatton: Done.

handshake_messages.go
```diff
@@ -1530,6 +1568,17 @@ func (m *certificateMsg13) marshal() (x []byte) {
 			sctLenPos[2] = uint8(sctLen >> 8)
 			sctLenPos[3] = uint8(sctLen)
 		}
+		if len(cert.dc) != 0 {
+			z[0] = byte(extensionDelegatedCredential >> 8)
```
henrydcase: Use "encoding/binary"

handshake_messages.go
henrydcase: Minimal PR is good idea. This should be done in 2 seprated commits:
1. https://github.com/cloudflare/tls-tris/pull/96
2. https://github.com/cloudflare/tls-tris/pull/97

Thanks for noticing

handshake_client.go
```diff
@@ -411,6 +414,48 @@ func (hs *clientHandshakeState) processCertsFromServer(certificates [][]byte) er
...
+
+	if cred != nil && !c.config.InsecureSkipVerify {
+		var now time.Time
+		if hs.c.config.Time != nil {
+			now = hs.c.config.Time()
+		} else {
+			now = time.Now()
+		}
+
+		if v, err := cred.Validate(c.peerCertificates[0], hs.c.vers, now); err != nil {
```
henrydcase: ``var err`` is already above, no need for ``:=``

handshake_client.go
```diff
@@ -411,6 +414,48 @@ func (hs *clientHandshakeState) processCertsFromServer(certificates [][]byte) er
...
+			c.sendAlert(alertIllegalParameter)
+			return fmt.Errorf("delegated_credential: %s", err)
+		} else if !v {
+			c.sendAlert(alertIllegalParameter)
```
henrydcase: Maybe nit-picking, but still - is "alertIllegalParameter" correct here?
I mean - I'm aware of 7 days validity "rule" for DC. But what if it doesn't verify correctly? Shouldn't it be "bad_certificate" as for TLS 1.3?

generate_cert.go
Bren2010: Ok, then just keep it in mind in case you're asked to upstream

cjpatton: Done. I'm adding a NOTE here about this at this line.

handshake_messages.go
cjpatton: Ack. See my comment on PR #96.

13.go
```diff
@@ -402,6 +402,15 @@ func (hs *serverHandshakeState) sendCertificate13() error {
 	if len(certEntries) > 0 && hs.clientHello.scts {
 		certEntries[0].sctList = hs.cert.SignedCertificateTimestamps
 	}
+
+	// If hs.dc is set (see hs.readClientHello()) then the server is using the
+	// delegated_credential extension. In TLS 1.3, the DC is added as an
+	// extension to the last CertificateEntry of Certificate.certficate_list
+	// (see ttps://tools.ietf.org/html/draft-ietf-tls-subcerts).
```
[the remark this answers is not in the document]
cjpatton: Done.

13.go
```diff
@@ -402,6 +402,15 @@ func (hs *serverHandshakeState) sendCertificate13() error {
 	if len(certEntries) > 0 && hs.clientHello.scts {
 		certEntries[0].sctList = hs.cert.SignedCertificateTimestamps
 	}
+
+	// If hs.dc is set (see hs.readClientHello()) then the server is using the
+	// delegated_credential extension. In TLS 1.3, the DC is added as an
+	// extension to the last CertificateEntry of Certificate.certficate_list
```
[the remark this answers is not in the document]
cjpatton: Done.

13.go
```diff
@@ -1032,12 +1041,21 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
...
 	hs.keySchedule.write(certMsg.marshal())
+
 	// Validate certificates.
 	certs := getCertsFromEntries(certMsg.certificates)
 	if err := hs.processCertsFromServer(certs); err != nil {
 		return err
 	}
 
+	// Validate the delegated credential if present.
+	if len(certMsg.certificates) > 0 {
```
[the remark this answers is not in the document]
cjpatton: The current is condition is necessary, since `len(certMsg.certificates)` could be `0` at this point. Verifying that the client indicated DCs is checked by `processDelegatedCredentialFromServer()`. I did it this way because this check is needed here and at `handshake_client.go:482`.

Added a comment clarifying this behavior.

handshake_messages.go
```diff
@@ -407,6 +412,11 @@ func (m *clientHelloMsg) marshal() []byte {
 		z[1] = byte(extensionEarlyData)
 		z = z[4:]
 	}
+	if m.delegatedCredential {
+		z[0] = byte(extensionDelegatedCredential >> 8)
```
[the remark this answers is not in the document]
cjpatton: Thank. GOD. :1st_place_medal: 

Done.

handshake_messages.go
```diff
@@ -976,7 +996,15 @@ func (m *serverHelloMsg) marshal() []byte {
 			z = z[len(sct)+2:]
 		}
 	}
-
+	if dcLen := len(m.dc); dcLen > 0 && m.vers == VersionTLS12 {
+		z[0] = byte(extensionDelegatedCredential >> 8)
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_client.go
cjpatton: But `v` is undefined, hence the need for `:=`.

handshake_client.go
cjpatton: This is what the spec says to do. I can go either way. Do you think the spec should be changed?

cjpatton:
@Bren2010 and @henrydcase, back to you!

handshake_client.go
henrydcase: okis

handshake_client.go
henrydcase: Hmmm... which place in the spec do you refer to exactly?

13.go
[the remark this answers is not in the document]
henrydcase: Right. I ment that this condition shouldn't be entered when AcceptDeletegatedCredential is set to false (so both len()>0 and AcceptDeletegatedCredential=true). But indeed, that's done.  All good here

handshake_messages.go
cjpatton: Done.

13.go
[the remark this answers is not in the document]
cjpatton: Ack.

handshake_client.go
cjpatton: Actually, this change *did* land: https://github.com/tlswg/tls-subcerts/commit/f249f76fd2c34ccb526fcdfc3d296da329c6d510

handshake_client.go
cjpatton: Done and done. Did `s/delegated_credential/delegated credential/g` here and below.

conn.go
```diff
@@ -47,17 +47,25 @@ type Conn struct {
...
+	ocspResponse []byte   // stapled OCSP response
+	scts         [][]byte // Signed certificate timestamps from server
+
 	peerCertificates []*x509.Certificate
+
 	// verifiedChains contains the certificate chains that we built, as
 	// opposed to the ones presented by the server.
 	verifiedChains [][]*x509.Certificate
+
+	// validatedD is set by a client who negotiates the use of a valid delegated
```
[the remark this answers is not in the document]
cjpatton: Done.

13.go
```diff
@@ -1032,12 +1042,23 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
...
 	certs := getCertsFromEntries(certMsg.certificates)
 	if err := hs.processCertsFromServer(certs); err != nil {
 		return err
 	}
 
+	// Validate the DC if present. The DC is is only processed if the
+	// extension was indicated by the ClientHello; otherwise this call will
+	// result in an "illegal_parameter" alert.
+	if len(certMsg.certificates) > 0 {
+		if err := hs.processDelegatedCredentialFromServer(
```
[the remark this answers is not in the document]
cjpatton: This is consistent with the call to `hs.processCertsFromServer()` above, as well as in `handshake_cleint.go`. Shall I fix all of these as you suggest?

handshake_client.go
```diff
@@ -431,6 +476,12 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		if err := hs.processCertsFromServer(certMsg.certificates); err != nil {
 			return err
 		}
+		// VAlidate the DC if present. The DC is is only processed if the
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_client.go
```diff
@@ -476,10 +527,23 @@ func (hs *clientHandshakeState) doFullHandshake() error {
...
+
+		// If the delegated_credential extension has successfully been
+		// negotiated, then the ServerKeyExchange (1.2) or CertificateVerify
+		// (1.3) signature will have been produced with the
+		// DelegatedCredential's private key.
+		if c.validatedDc != nil {
+			hs.peerPublicKey = c.validatedDc.PublicKey
+		}
+
+		err = keyAgreement.processServerKeyExchange(
```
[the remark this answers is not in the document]
cjpatton: Removed wrap here.

handshake_client.go
```diff
@@ -528,7 +592,8 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		}
 	}
 
-	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(c.config, hs.hello, c.peerCertificates[0])
+	preMasterSecret, ckx, err := keyAgreement.generateClientKeyExchange(
```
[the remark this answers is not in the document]
cjpatton: Removed wrap here.

conn.go
```diff
@@ -47,17 +47,25 @@ type Conn struct {
...
-	didResume        bool // whether this connection was a session resumption
-	cipherSuite      uint16
-	ocspResponse     []byte   // stapled OCSP response
-	scts             [][]byte // signed certificate timestamps from server
+	handshakes   int
+	didResume    bool // whether this connection was a session resumption
+	cipherSuite  uint16
+	ocspResponse []byte   // stapled OCSP response
+	scts         [][]byte // Signed certificate timestamps from server
+
```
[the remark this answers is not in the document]
cjpatton: You want me to remove these newlines, or add more?

handshake_messages.go
```diff
@@ -758,6 +772,7 @@ type serverHelloMsg struct {
 	secureRenegotiation          []byte
 	secureRenegotiationSupported bool
 	alpnProtocol                 string
+	dc                           []byte
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_server.go
```diff
@@ -30,6 +30,11 @@ type serverHandshakeState struct {
 	clientHello           *clientHelloMsg
 	hello                 *serverHelloMsg
 	cert                  *Certificate
+	privateKey            crypto.PrivateKey
+
+	// A marshaled DeleegatedCredential to be sent to the client in the
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_client.go
```diff
@@ -18,15 +18,17 @@ import (
...
-	suite        *cipherSuite
-	masterSecret []byte
-	session      *ClientSessionState
+	c             *Conn
+	serverHello   *serverHelloMsg
+	hello         *clientHelloMsg
+	suite         *cipherSuite
+	masterSecret  []byte
+	session       *ClientSessionState
+	peerPublicKey crypto.PublicKey
```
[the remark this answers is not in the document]
cjpatton: The `validatedDc` field is analogous to the `verifiedCerts` field set by the cert verification code in `processCertsFromServer()`. Right now the only place that `validatedDc` is used after it's set is by the tests in `subcerts_test.go`. It indicates whether the DCs were used in the handshake.

handshake_client.go
```diff
@@ -431,6 +476,12 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		if err := hs.processCertsFromServer(certMsg.certificates); err != nil {
 			return err
 		}
+		// VAlidate the DC if present. The DC is is only processed if the
+		// extension was indicated by the ClientHello; otherwise this call will
+		// result in an "illegal_parameter" alert.
+		if err := hs.processDelegatedCredentialFromServer(hs.serverHello.dc); err != nil {
```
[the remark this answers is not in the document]
cjpatton: Again, this change would break consistency with the surrounding code. (See the call to `processCertsFromServer()` above.)

handshake_server.go
```diff
@@ -30,6 +30,11 @@ type serverHandshakeState struct {
 	clientHello           *clientHelloMsg
 	hello                 *serverHelloMsg
 	cert                  *Certificate
+	privateKey            crypto.PrivateKey
```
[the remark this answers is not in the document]
cjpatton: It's called by `doFullHandshake()`, too.

cjpatton:
I'll clean up the commit history once we're done reviewing. In particular, I'll collapse all commits following (and including) https://github.com/cloudflare/tls-tris/pull/95/commits/7e2f344052e47156a73d11c951f20fc411e3cae4 into one.

13.go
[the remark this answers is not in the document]
cjpatton: (Note that would make the PR non-minimal.)

handshake_client.go
cjpatton: Donel.

handshake_server.go
```diff
@@ -294,16 +299,54 @@ Curves:
 		}
 	}
 
-	hs.cert, err = c.config.getCertificate(hs.clientHelloInfo())
+	ch := hs.clientHelloInfo()
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_server.go
```diff
@@ -294,16 +299,54 @@ Curves:
...
+				dc.Raw, err = dc.Marshal()
+				if err != nil {
+					c.sendAlert(alertInternalError)
+					return false, err
+				}
+			}
+			hs.delegatedCredential = dc.Raw
+
+			// For TLS 1.2, the DC is an extension to the ServerHello.
+			if c.vers == VersionTLS12 && hs.hello != nil {
```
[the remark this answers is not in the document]
cjpatton: Agreed. Removed `&& hs.hello != nil` here and above.

subcerts.go
```diff
@@ -0,0 +1,534 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tls
+
+// Delegated credentials for TLS
+// (https://tools.ietf.org/html/draft-ietf-tls-subcerts) is an IETF Internet
+// draft and proposed TLS extension. If the client supports this extension, then
+// the servery may use a "delegated credential" as the signing key in the
```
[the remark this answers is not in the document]

subcerts.go
```diff
@@ -0,0 +1,534 @@
...
+func CreateDelegationUsagePKIXExtension() *pkix.Extension {
+	return &pkix.Extension{
+		Id:       delegationUsageId,
+		Critical: false,
+		Value:    nil,
+	}
+}
+
+// canDelegate checks that a certificate can be used for delegated credentials
+// and reuturns nil if it can. Otherwise it returns an error conveying the
```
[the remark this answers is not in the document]
cjpatton: Nice catch. Done.

subcerts.go
```diff
@@ -0,0 +1,534 @@
...
+
+		pk, ok := cert.PublicKey.(*ecdsa.PublicKey)
+		if !ok {
+			return false, errors.New("expected ECDSA public key")
+		}
+		sig := new(ecdsaSignature)
+		if _, err = asn1.Unmarshal(dc.Signature, sig); err != nil {
+			return false, err
+		}
+		if sig.R.Sign() <= 0 || sig.S.Sign() <= 0 {
```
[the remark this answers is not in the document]
cjpatton: Huh, you're right, but this check is done in `verifyHandshakeSignature()`, which is why I did it here. 

Done.

subcerts.go
```diff
@@ -0,0 +1,534 @@
...
+
+	h := hash.New()
+
+	// The header.
+	h.Write(bytes.Repeat([]byte{0x20}, 64))
+	h.Write([]byte("TLS, server delegated credentials"))
+	h.Write([]byte{0x00})
+
+	// The protocol version.
+	serializedVer := make([]byte, 2)
```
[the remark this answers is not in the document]

cipher_suites.go
```diff
@@ -26,15 +26,15 @@ type keyAgreement interface {
 	// In the case that the key agreement protocol doesn't use a
 	// ServerKeyExchange message, generateServerKeyExchange can return nil,
 	// nil.
-	generateServerKeyExchange(*Config, *Certificate, *clientHelloMsg, *serverHelloMsg) (*serverKeyExchangeMsg, error)
```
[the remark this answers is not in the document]
cjpatton: Do you mean a separate PR? If you'd like I can create a PR with just these changes, then rebase this PR.

It's certainly sufficient to pass in the whole `tls.Certificate` (resp.`x509.Certificate`), but all that's needed by the key agreement functions is the private key (resp. public key). Refactoring the code in this way was necessary for baking in the DC stuff.

handshake_server.go
```diff
@@ -299,11 +304,48 @@ Curves:
...
+	// Set the private key for this handshake to the certificate's secret key.
+	hs.privateKey = hs.cert.PrivateKey
+
+	if hs.clientHello.scts {
 		hs.hello.scts = hs.cert.SignedCertificateTimestamps
 	}
 
-	if priv, ok := hs.cert.PrivateKey.(crypto.Signer); ok {
+	// Set the private key to the DC private key if the client and server are
+	// willing to negotiate the deleaged_credential extension.
```
Bren2010: delegated credential

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+
+		serializedPublicKey, err := x509.MarshalPKIXPublicKey(cred.publicKey)
+		if err != nil {
+			return nil, err
+		}
+		return serializedPublicKey, nil
+
+	default:
+		return nil, fmt.Errorf(
+			"unsupported signature scheme: 0x%04x", cred.scheme)
```
Bren2010: unwrap

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+}
+
+// unmarshalSubjectPublicKeyInfo parses a DER encoded SubjectPublicKeyINfo
+// structure into a public key and its corresponding algorithm.
+func unmarshalSubjectPublicKeyInfo(
+	serialized []byte) (crypto.PublicKey, SignatureScheme, error) {
```
Bren2010: unwrap, no empty newline immediately below the funciton signature, INfo -> Info

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+
+	publicKey, err := x509.ParsePKIXPublicKey(serialized)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	switch pk := publicKey.(type) {
+	case *ecdsa.PublicKey:
+
```
Bren2010: no newline

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+			return pk, ECDSAWithP384AndSHA384, nil
+		} else if curveName == "P-521" {
+			return pk, ECDSAWithP521AndSHA512, nil
+		} else {
+			return nil, 0, fmt.Errorf("curve %s s not supported", curveName)
+		}
+
+	default:
+		return nil, 0, fmt.Errorf(
+			"unsupported delgation key type: %s", reflect.TypeOf(pk))
```
Bren2010: unwrap, fmt.Errorf("... %T", pk) so you don't need to import reflect

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+		if extension.Id.Equal(delegationUsageId) {
+			return true
+		}
+	}
+	return false
+}
+
+// This structure stores the public components of a credential.
+type credential struct {
+	validTime time.Duration
```
Bren2010: Call this something more descriptive like `notAfter`? Also, where is the signature over the DC?

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+	// Write the valid_time field.
+	serialized := make([]byte, 6)
+	binary.BigEndian.PutUint32(serialized, uint32(cred.validTime/time.Second))
+
+	// Encode the public key.
+	serializedPublicKey, err := cred.marshalSubjectPublicKeyInfo()
+	if err != nil {
+		return nil, err
+	}
+
```
Bren2010: no newline, as this flows directly into the next statement

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+	// Write the length of the public_key field.
+	binary.BigEndian.PutUint16(serialized[4:], uint16(len(serializedPublicKey)))
+
+	// Write the public key.
+	return append(serialized, serializedPublicKey...), nil
+}
+
+// unmarshalCredential decodes a credential and returns it.
+func unmarshalCredential(serialized []byte) (*credential, error) {
+
```
Bren2010: no newline

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+	// Time in nonsecnds for which the credential is valid. When this data
+	// structure is serialized, this value is converted to a uint32 representing
+	// the duration in seconds.
+	ValidTime time.Duration
+
+	// The credential public key.
+	PublicKey crypto.PublicKey
+
+	// The credential prviate key.
```
Bren2010: private

subcerts.go
```diff
@@ -0,0 +1,533 @@
...
+
+// NewDelegatedCredential creates a new delegated credential using `cert` for
+// delegation. It generates a public/private key pair for the provided signature
+// algorithm (`scheme`), validity interval (defined by `cert.Leaf.notBefore` and
+// `validTime`), and TLS version (`vers`), and signs it using `cert.PrivateKey`.
+func NewDelegatedCredential(
+	cert *Certificate,
+	scheme SignatureScheme,
+	validTime time.Duration,
+	vers uint16) (*DelegatedCredential, error) {
```
Bren2010: no wrap

subcerts.go
Bren2010: oh, it's a duration. `validFor: 3*time.Second` ?

conn.go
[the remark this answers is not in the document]
cjpatton: Done.

handshake_client.go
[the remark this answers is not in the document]
cjpatton: Done. I named the variable `pk` instead of `peerPublicKey`.

handshake_client.go
```diff
@@ -411,6 +414,48 @@ func (hs *clientHandshakeState) processCertsFromServer(certificates [][]byte) er
...
+
+		cred, err = unmarshalDelegatedCredential(dc)
+		if err != nil {
+			c.sendAlert(alertDecodeError)
+			return fmt.Errorf("tls: delegated credential: %s", err)
+		}
+	}
+
+	if cred != nil && !c.config.InsecureSkipVerify {
+		var now time.Time
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_server.go
cjpatton: Done.

subcerts.go
cjpatton: `validTime` matches the field name in the spec.

subcerts.go

subcerts.go

subcerts.go

subcerts.go

subcerts.go

subcerts.go

subcerts.go

subcerts.go

cipher_suites.go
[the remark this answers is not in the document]
henrydcase: separate commit in this PR would be enough

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+
+// unmarshalCredential decodes a credential and returns it.
+func unmarshalCredential(serialized []byte) (*credential, error) {
+	// Check that the serialized credential is long enough and compute the
+	// length of the serialized public key.
+	serializedPublicKeyLen, err := getCredentialLen(serialized)
+	if err != nil {
+		return nil, err
+	}
+	serializedPublicKeyLen -= 6
```
henrydcase: unused

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+
+	// Parse the validity time.
+	validTime := time.Duration(binary.BigEndian.Uint32(serialized)) * time.Second
+
+	// Parse the SubjectPublicKeyInfo.
+	serialized = serialized[6:]
```
henrydcase: comment why '6'. can be in comment for 'getCredentialLen'

cipher_suites.go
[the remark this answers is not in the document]
cjpatton: Done.

subcerts_test.go
```diff
@@ -0,0 +1,453 @@
...
+
+// These test keys were generated with the following program, available in the
+// crypto/tls directory:
+//
+//		go run generate_cert.go -ecdsa-curve P256 -host 127.0.0.1 -dc
+//
+// To get a certificate without the DelegationUsage extension, remove the `-dc`
+// parameter.
+
+var delegatorCertPEM = `-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
cjpatton: The current convention is consistent with the rest of the code base, see `tls_test.go`. Shall I break  the consistency?

subcerts.go
cjpatton: Done. Nice catch.

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+
+	return 6 + serializedPublicKeyLen, nil
+}
+
+// DelegatedCredential stores a credential and its delegation.
+type DelegatedCredential struct {
+	// The serialized form of the credential.
+	Raw []byte
+
+	// Time in nonsecnds for which the credential is valid. When this data
```
Bren2010: ~~nanoseconds~~ Specifying unit shouldn't be necessary, b/c its a time.Duration. "ValidTime is how long the cred is valid"

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+	// structure is serialized, this value is converted to a uint32 representing
+	// the duration in seconds.
+	ValidTime time.Duration
+
+	// The credential public key.
+	PublicKey crypto.PublicKey
```
Bren2010: I think pretty much all private keys have the public key embedded, so this allows an illegal state (PublicKey != PrivateKey.PublicKey). Can it be removed?

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+// algorithm (`scheme`), validity interval (defined by `cert.Leaf.notBefore` and
+// `validTime`), and TLS version (`vers`), and signs it using `cert.PrivateKey`.
+func NewDelegatedCredential(cert *Certificate, scheme SignatureScheme, validTime time.Duration, vers uint16) (*DelegatedCredential, error) {
+	// The granularity of DC validity is seconds.
+	validTime = validTime.Round(time.Second)
+
+	// Parse the leaf certificate if needed.
+	var err error
+	if cert.Leaf == nil {
+		cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
```
Bren2010:
```
if len(cert.Certificate) == 0 {
    return err
}
```
?

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+	// Check that the leaf certificate can be used for delegation.
+	if !canDelegate(cert.Leaf) {
+		return nil, errNoDelegationUsage
+	}
+
+	// Extract the delegator signature scheme rom the certificate.
+	var delegatorScheme SignatureScheme
+	switch sk := cert.PrivateKey.(type) {
+	case *ecdsa.PrivateKey:
+
```
Bren2010: no newline

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+
+	// Generate a new key pair.
+	var sk crypto.PrivateKey
+	var pk crypto.PublicKey
+	switch scheme {
+	case ECDSAWithP256AndSHA256,
+		ECDSAWithP384AndSHA384,
+		ECDSAWithP521AndSHA512:
+
```
Bren2010: no newline

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+		sig, err = sk.Sign(rand.Reader, in, opts)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unsupported delgation key type: %T", sk)
+	}
+
+	return &DelegatedCredential{
+		nil, validTime, pk, sk, scheme, delegatorScheme, sig}, nil
```
Bren2010: Long-form would be good here. It handles the wrap problem, and makes what-is-what clearer to me,
```
&DelegatedCredential{
    ValidTime: validTime,
 ...
```

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+// IsExpired returns true if the credential has expired. The end of the validity
+// interval is defined as the deleagtor certificate's notBefore field (`start`)
```
Bren2010: delegator

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+
+	// Prepare the credential for digital signing.
+	hash := getHash(delegatorScheme)
+	cred := &credential{validTime, pk, scheme}
+	in, err := prepareDelegation(hash, cred, cert.Leaf.Raw, delegatorScheme, vers)
+	if err != nil {
+		return nil, err
+	}
+
+	// Sign the creential.
```
Bren2010: credential

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+		return false, err
+	}
+
+	// TODO(any) This code overlaps signficantly with verifyHandshakeSignature()
+	// in ../auth.go. This should be refactored.
+	switch dc.Scheme {
+	case ECDSAWithP256AndSHA256,
+		ECDSAWithP384AndSHA384,
+		ECDSAWithP521AndSHA512:
+
```
Bren2010: no newline

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+	if dc.IsExpired(cert.NotBefore, now) {
+		return false, errors.New("credential has expired")
+	}
+
+	if dc.InvalidTTL(cert.NotBefore, now) {
+		return false, errors.New("credential TTL is invalid")
+	}
+
+	// Prepare the crednetial for verification.
```
Bren2010: credential

subcerts.go
Bren2010: Also, a lot of this file is big if-tables and switch statements and feels repetitive. I would think about how to reduce this, or compartmentalize it. Not necessary to do, just a thought.

subcerts.go
```diff
@@ -0,0 +1,508 @@
...
+	binary.BigEndian.PutUint16(serializedSignature, uint16(len(dc.Signature)))
+	serializedSignature = append(serializedSignature, dc.Signature...)
+	serialized = append(serialized, serializedSignature...)
+
+	dc.Raw = serialized
+	return serialized, nil
+}
+
+// unmarshalDelegatedCredential decodes a DelegatedCredential structure.
+func unmarshalDelegatedCredential(serialized []byte) (*DelegatedCredential, error) {
```
Bren2010: If unmarshal is un-exported, how do I unmarshal a marshaled DC?

subcerts_test.go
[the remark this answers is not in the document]
henrydcase: I'll leave it up to you, not really important I guess

subcerts.go
```diff
@@ -0,0 +1,505 @@
...
+		return 0 // Unknown hash function
+	}
+}
+
+// prepareDelegation returns a hash of the message that the delegator is to
+// sign. The inputs are the credential (cred), the DER-encoded delegator
+// certificate (`delegatorCert`), the signature scheme of the delegator
+// (`delegatorScheme`), and the protocol version (`vers`) in which the credential
+// is to be used.
+func prepareDelegation(hash crypto.Hash, cred *credential, delegatorCert []byte, delegatorScheme SignatureScheme, vers uint16) ([]byte, error) {
```
Bren2010: prepareDelegation -> sumDelegation?

generate_cert.go
```diff
@@ -158,4 +171,5 @@ func main() {
 	pem.Encode(keyOut, pemBlockForKey(priv))
 	keyOut.Close()
 	log.Print("written key.pem\n")
+	0
```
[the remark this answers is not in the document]
cjpatton: Done.

handshake_client.go
```diff
@@ -411,6 +413,46 @@ func (hs *clientHandshakeState) processCertsFromServer(certificates [][]byte) er
...
+
+		cred, err = unmarshalDelegatedCredential(dc)
+		if err != nil {
+			c.sendAlert(alertDecodeError)
+			return fmt.Errorf("tls: delegated credential: %s", err)
+		}
+	}
+
+	if cred != nil && !c.config.InsecureSkipVerify {
+		now := time.Now()
```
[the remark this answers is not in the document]
cjpatton: Nice catch. Done.

subcerts.go
cjpatton: Removed "in nanoseconds".

subcerts.go
cjpatton: I would prefer not to, since the crypto.PrivateKey doesn't require one to implement a Public() function, like crypto.Signer does. I suppose I could change this to a crypto.Signer? The problem with that solution is that it would be inconsistent with tls.Certificate, which has a crypto.PrivateKey field.

subcerts.go
cjpatton: ^Ack.

cjpatton: ^^Agreed. Done.

subcerts.go

subcerts.go

subcerts.go

subcerts.go
cjpatton: You don't, I guess? Originally this function was exported, I'm happy to export this. My thinking was that we wanted to minimize our changes (i.e., additions) to the API as much as possible; since DCs are meant to only be consumed by TLS clients and servers, it made sense to not export it.

subcerts.go
cjpatton: prepareDelegation is consistent with prepareDigitallySigned, defined in 13.go.

subcerts_test.go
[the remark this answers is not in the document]
cjpatton: Ack. Leaving as is.

cjpatton:
A note about the API: One goal of this implementation is to minimize changes to the `crypto/tls` API. WRT the API, the changes are as follows:
 - Add `AcceptDelegatedCredential` to `tls.ClientHelloInfo`
 - Add `AcceptDelegatedCredential` and `GetDelegatedCredential` to `tls.Config`
 - Add structure `DelegatedCredential` and a constructor for it

I claim that these changes are necessary. But they might not be sufficient for every application. In particular, this is fine if the DCs are generated and signed on the same machine; but there's no way for someone to generate the DC and get it signed on a different machine. For this we would need some additional changes:
 - Add structure `Credential` and a constructor for it. (Currently there's an unexported `credential` structure used internally by `subcerts.go`; we would simply export this.)
 - Add function `Delegate` to `tls.Certificate`. This takes as input a `Credential` and outputs a `DelegatedCredential`. 

So my question is this: I could make this change in this PR, or in a different PR that follows it. Do @Bren2010  or @henrydcase have a preference?

henrydcase:
Different PR I would say. But also I would not hurry too much with defining/freezing API at this stage. My preference would be to export only those functions/flags which are absolutelly necesairy. We can always export additional functions if/when needed, but removing/changing is much harder.

cjpatton:
Sounds good to me. @Bren2010, any additional comments? Otherwise I'll hand it to @grittygrease for review.

subcerts.go
Bren2010: Not sure I understand. My problem is that you're reaching into a slice without checking bounds

cjpatton: Gotcha. Done.

subcerts.go
```diff
@@ -0,0 +1,515 @@
...
+
+	default:
+		return false, fmt.Errorf(
+			"unsupported signature scheme: 0x%04x", dc.Scheme)
+	}
+}
+
+// Marshal encodes a DelegatedCredential structure per the spec. It also sets
+// dc.Raw to the output as a side effect.
+func (dc *DelegatedCredential) Marshal() ([]byte, error) {
```
Bren2010: crypto/tls doesn't provide any other marshalers / unmarshalers. Why does it make sense to have these exported?

common.go
```diff
@@ -607,6 +612,22 @@ type Config struct {
...
+	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
+	AcceptDelegatedCredential bool
+
+	// GetDelegatedCredential returns a DelegatedCredential for use with the
+	// delegated credential extension based on the ClientHello and TLS version
+	// selected for the session. If this is nil, then the server will not offer
+	// a DelegatedCredential.
+	//
+	// This value has no meaning for the client.
+	GetDelegatedCredential func(*ClientHelloInfo, uint16) (*DelegatedCredential, error)
```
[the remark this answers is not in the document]
cjpatton: This is necessary because the DC is bound to the TLS version in which it's to be used. `ClientHelloInfo` provides the range of version supported by the client, but doesn't reflect the server's decision. Passing in the version here allows us to choose the DC after the server has picked the DC.

handshake_client.go
```diff
@@ -476,10 +518,22 @@ func (hs *clientHandshakeState) doFullHandshake() error {
...
 		hs.finishedHash.Write(skx.marshal())
-		err = keyAgreement.processServerKeyExchange(c.config, hs.hello, hs.serverHello, c.peerCertificates[0], skx)
+
+		// If the delegated credential extension has successfully been
+		// negotiated, then the ServerKeyExchange (1.2) or CertificateVerify
+		// (1.3) signature will have been produced with the
+		// DelegatedCredential's private key.
+		if c.verifiedDc != nil {
+			pk = c.verifiedDc.PublicKey
+		}
```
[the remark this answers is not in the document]
cjpatton: The short answser: the DC spec doesn't support KEM-based key exchange.

common.go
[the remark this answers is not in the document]
Bren2010: Hm. Do you know the purpose of binding to the version?

I think there was a GREASE-releated project where every few days, Chrome and Cloudflare's edge start speaking a new version of TLS. This puts a wrench in that.

handshake_client.go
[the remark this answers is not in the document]
Bren2010: I think one of these need to happen:

- Move this `if` up out of the nested `if` to where pk is defined. This is enough to support KEM-based key exchanges?
- Somewhere we enforce that client doesn't offer KEM-based kex with DC, and server rejects it with an error.

Right now, the client and server both walk a valid code-path and come to a different understanding of the server's public key.

handshake_server.go
```diff
@@ -299,11 +304,48 @@ Curves:
...
+	// private key to the DC private key.
+	if c.config.GetDelegatedCredential != nil && hs.clientHello.delegatedCredential {
+		dc, err := c.config.GetDelegatedCredential(hs.clientHelloInfo(), c.vers)
+		if err != nil {
+			c.sendAlert(alertInternalError)
+			return false, err
+		}
+
+		// Set the handshake private key.
+		if dc != nil && c.vers >= VersionTLS12 {
```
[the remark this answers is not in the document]
Bren2010: What is the purpose of fetching the DC if you're never going to use it because `c.vers < 1.2` ?

subcerts.go
Bren2010: Why does somebody outside of crypto/tls need to be able to compute the TLS wire-format DC?

common.go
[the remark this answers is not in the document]
cjpatton: But GREASE doesn't change the TLS version, does it? It just offers random extensions and ciphersuites.

Here's what the spec says. Section 2.1: "Delegated credentials allow signed messages to be bound to specific versions of TLS. This prevents them from being used for other protocols if a service owner allows multiple versions of TLS." Section 4: "It is bound to the protocol version that is negotiated. This is intended to avoid cross-protocol attacks with signing oracles."

handshake_client.go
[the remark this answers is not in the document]
cjpatton: Nice catch. Moved the inner `if` to before the outer `if`, just after the point where `pk` is defined.

handshake_server.go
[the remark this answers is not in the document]
cjpatton: There's no point at all. I misunderstood your comment the first time. Moved `c.vers >= 1.2` to outer `if` statement.

subcerts.go
cjpatton: We need to be able to use these APIs from Keyless, for example.

subcerts.go
```diff
@@ -0,0 +1,512 @@
...
+			return nil, err
+		}
+	}
+
+	// Check that the leaf certificate can be used for delegation.
+	if !canDelegate(cert.Leaf) {
+		return nil, errNoDelegationUsage
+	}
+
+	// Extract the delegator signature scheme rom the certificate.
```
henrydcase: "rom" -> form

handshake_server.go
```diff
@@ -299,11 +304,48 @@ Curves:
...
 
-	if priv, ok := hs.cert.PrivateKey.(crypto.Signer); ok {
+	// Set the private key to the DC private key if the client and server are
+	// willing to negotiate the delegated credential extension.
+	//
+	// Check to see if a DelegatedCredential is available and should be used.
+	// If one is available, the session is using TLS >= 1.2, and the client
+	// accepts the delegated credential extension, then set the handshake
+	// private key to the DC private key.
+	if c.config.GetDelegatedCredential != nil && hs.clientHello.delegatedCredential {
```
henrydcase: I think there is an update needed for the documentation and TLS 1.3. Spec says
```
   The DelegatedCredential contains a signature from the public key in
   the end-entity certificate using a signature algorithm advertised by
   the client in the "signature_algorithms" extension.  Additionally,
   the credential's public key MUST be of a type that enables at least
   one of the supported signature algorithms.  A Delegated credential
   MUST NOT be negotiated by the server if its signature is not
   compatible with any of the supported signature algorithms or the
   credential's public key is not usable with the supported signature
   algorithms of the client, even if the client advertises support for
   delegated credentials.
```

So if delegated credential returns signature type which is different than server public key type or is not on a ``signature_algorithms`` list, then it is a bug.
But TLS 1.3 introduces new extension called ``signature_algotirhms_cert`` (see 4.2.3 of a spec). It is a list of algorithms that client is able to verify. I think this implementation should look at ``signature_algotirhms_list`` if it is available and ``signature_algorithms`` in the other case.

handshake_client.go
```diff
@@ -624,6 +678,12 @@ func (hs *clientHandshakeState) serverResumedSession() bool {
 func (hs *clientHandshakeState) processServerHello() (bool, error) {
 	c := hs.c
 
+	// Process the delegated credential extension.
+	if hs.serverHello.delegatedCredential != nil && hs.serverHello.vers != VersionTLS12 {
```
henrydcase: can you move this check to ``processDelegatedCredentialFromServer`` ?

subcerts.go

13.go
```diff
@@ -1032,12 +1042,23 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
...
+	// Validate the DC if present. The DC is only processed if the extension was
+	// indicated by the ClientHello; otherwise this call will result in an
+	// "illegal_parameter" alert.
+	if len(certMsg.certificates) > 0 {
+		if err := hs.processDelegatedCredentialFromServer(
+			certMsg.certificates[0].delegatedCredential); err != nil {
+			return err
+		}
+	}
+
```
henrydcase: I think there is a bug in case TLSv1.3 is used.

Chapter 3 of the spaec says:
```
   On receiving a credential and a certificate chain, the client
   validates the certificate chain and matches the end-entity
   certificate to the server's expected identity following its normal
   procedures.  It then takes the following additional steps:
```
This part is done indeed, but then it says:

```
   o  Use the public key in the credential to verify a signature
      provided in the handshake.  That is the CertificateVerify message
      in TLS 1.3 or ...

```
I don't think it is done. In order to do that, you would need to modify call to ``verifyPeerCertificate()``. Currently call to this function takes ``hs.c.peerCertificate[0].PublicKey``, which is not a delegated credential key (except if I'm missing something).

BTW:  Better to merge #104 before (name of this function is missleading)

handshake_server.go
cjpatton: I'm not sure I understand. A correct implementation of `GetDelegatedCredential` will return a DC that uses a signature algorithm listed in `ClientHelloInfo.SignatureSchemes`. The question, then, is how is this list created. Per draft >= 23, this must be same as the list in the `signature_algorithms` extension.

Here's where I don't follow. The `signature_algorithms_cert` extension specifies the list of algorithms the client supports _for the purpose of verifying the server's cert chain._ If the client specifies this extension, then the `signature_algorithms` list applies only to verifying the handshake. (If it unspecified, then `signature_algorithms` applies to both.) So I don't see how there could be a bug here.

Unless the semantics of `ClientHelloInfo.SignatureSchemes` is changing in #102. Is that the case?

handshake_client.go
cjpatton: `processDelegatedCredentialFromServer` is called from `doTLS13Handshake` (13.go:1056), so if I move the check there, the condition must be != TLS 1.2. Shall I move it, or change the condition on handshake_client.go:682?

13.go
cjpatton: Yup, for TLS 1.3 it looks like I forgot to change the key used to sign/verify the handshake when using a DC. The next commit fixes this and adds a test case. Thanks for catching this.

handshake_client.go
henrydcase: I think you should move it and change.
It wouldn't be good if you fail it for TLS 1.3 I think

handshake_server.go
henrydcase: So, in TLS 1.3:
* ``signature_algorithms_cert``: list of algorithms client is able to use to verify certificates (or server in case of client verification)
* ``signature_algorithms``: list of algorithms client is able to use to verify handshake signature. More specifically this list is used for CertificateVerify message only ...
* ... except if - ``signature_algorithms_cert`` is not provided  - then ``signature_algorithms`` list is used

cjpatton: Right. So the semantics of `ClientHelloInfo.SignatureSchemes` in #102 is wrong. Effectively, there is one set of signature schemes the server uses to pick its certificate. If the `signature_algorithms_cert` extension is provided, then that's the set; otherwise the set provided in `signature_algorithms` is the set.

henrydcase: Nop, code sets SignatureSchemes to list from which server needs to choose ceritficates. If TLS version is 1.3 AND ``signature_algorithms_cert`` is provided ( which corresponds to ``hs.clientHello.supportedSignatureAlgorithmsCert`` variable), then this set will be used. Otherwise ``hs.clientHello.supportedSignatureAlgorithms``

handshake_client.go
cjpatton: In TLS 1.3, the DC is returned in the Certificate message, not the ServerHello.

subcerts.go
```diff
@@ -0,0 +1,515 @@
...
+	// Time for which the credential is valid. When this data structure is
+	// serialized, this value is converted to a uint32 representing the duration
+	// in seconds.
+	ValidTime time.Duration
+
+	// The credential public key.
+	PublicKey crypto.PublicKey
+
+	// The credential private key.
+	PrivateKey crypto.PrivateKey
```
Bren2010: I think we agreed after talking irl that this field would be removed and returned by NewDC. It's awkward having a field that is silently elided by Marshal -> Unmarshal.

handshake_client.go
```diff
@@ -431,6 +467,12 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		if err := hs.processCertsFromServer(certMsg.certificates); err != nil {
 			return err
 		}
+		// Validate the DC if present. The DC is is only processed if the
```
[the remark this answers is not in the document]
cjpatton: Done.

subcerts.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+// DelegatedCredential stores a credential and its delegation.
+type DelegatedCredential struct {
+	// The serialized form of the credential.
+	Raw []byte
+
+	// Time for which the credential is valid. When this data structure is
+	// serialized, this value is converted to a uint32 representing the duration
+	// in seconds.
```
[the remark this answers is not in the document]
cjpatton: Revised the comment.

handshake_server.go
cjpatton: See comment thread in #102.

handshake_client.go
henrydcase: In such case:
* ``!=`` is OK, still please move the code
* comment is wrong - you don't do any processing of DC here
* in ``serverHelloMsg`` struct you need a comment explaining that ``delegatedCredential`` can be set for TLS 1.2 only

cjpatton: Gotcha, I'm on it.

cjpatton: Done.

subcerts.go
cjpatton: This was removed by https://github.com/cloudflare/tls-tris/pull/95/commits/54ec0a4b3583d4e9b0c5360917aca820b576bb9f

cjpatton: (That commit is gone now; I smashed all of the changes since review began into a new commit https://github.com/cloudflare/tls-tris/pull/95/commits/a34d64ca3f73b18fa9858ec8f7ad7053fd16aadd.)

cjpatton:
Hi @Bren2010 and @henrydcase, I squished the commits into just two: one that refactors keyAgreement (as requested by @henrydcase) and another that implements DCs.

henrydcase:
This PR must be merge after #102

henrydcase:
I actually would like to have interoperability tested by interoperability tests (``_dev/interop_test_runner``). Can you think about it and propose a solution?

cjpatton:
That'll require DC support on the BSSL server, right? We're not quite there yet.

handshake_client.go
```diff
@@ -477,10 +529,20 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 
 	keyAgreement := hs.suite.ka(c.vers)
 
+	// Set the public key used to verify the handshake.
+	pk := c.peerCertificates[0].PublicKey
```
henrydcase: Quick question - I havn't looked enough in to code but I'm wondering why we can't reuse GetCertificate and VerifyPeerCertificate to get/validate credentials.
I know that those functions take/return "Certificate" object and you are using PublicKey and PrivateKey objects internally. But isn't it possible to wrap those keys (and other attributes) it in Certificate object? I think integration would be a bit easier then.

handshake_server.go
```diff
@@ -299,11 +304,48 @@ Curves:
...
 
-	if priv, ok := hs.cert.PrivateKey.(crypto.Signer); ok {
+	// Set the private key to the DC private key if the client and server are
+	// willing to negotiate the delegated credential extension.
+	//
+	// Check to see if a DelegatedCredential is available and should be used.
+	// If one is available, the session is using TLS >= 1.2, and the client
+	// accepts the delegated credential extension, then set the handshake
+	// private key to the DC private key.
+	if c.config.GetDelegatedCredential != nil && hs.clientHello.delegatedCredential && c.vers >= VersionTLS12 {
```
henrydcase: Another quick question - "GetConfigForClient" this can't be reused here?

handshake_client.go
cjpatton: I tried this originally but decided against. I don't exactly remember why it didn't work, but one reason is that choosing the DC requires knowing what TLS version has been negotiated. This can't be determined given only the parameters of `GetCertificate`.

henrydcase: Seems to me that GetCertificate is a method of "ClientHelloInfo", which has a "Conn" which stores TLS version. But anyway, if you said you checked it, then it's probably not worth trying.

handshake_server.go
cjpatton: This could be an option, but again, the TLS version chosen by the server can't be determined from `ClientHelloInfo` (the input to `GetConfigForClient`). This is needed because the delegation binds the DC to the protocol.

One way we could get around this is by having an internal `map[uint16]*DelegatedCredential` field of the serverHandshakeState structure. But we'd have to expose an interface for populating this. I thought it would be better to leave how DCs are chosen to the deployment, hence the `GetDelegatedCredential` callback. This also aligns nicely with `GetCertificate`.

handshake_client.go
cjpatton: Aw yeah you're right. But does crypto's Conn epxort the version?

henrydcase: c.ver - it's used all over the code

cjpatton: Yeah it doesn't export the version. I remember I looked at this a couple weeks ago.

cjpatton: But if you're writing your own `GetCertificate`, you can't access ClientHelloInfo.Conn.vers.

cjpatton: (Outside of crypto/tls)

henrydcase: Oh. it's private field, but that's not  a problem at all. Simply adding public method will be both enough and actually useful

cjpatton: We could add GetVersion() to tls.Conn. But the field in ClientHelloInfo is a net.Conn: Do you think this could be an issue?

cjpatton:
The last change just moves a NOTE from generate_cert.go to subcerts.go. Feel free to merge anytime.

cjpatton:
Actually, before you merge, let me update the OID. Currently it's 2.5.29.99, which we don't have the right to assign. I'm changing it to 1.3.6.1.4.1.44363.44.

cjpatton:
Alright, LGTM.

PR #96, opened 2018-06-19
Applies go fmt to all the code

henrydcase:
Simply go fmt all the thingz.
``` go fmt ./... ```

No functional changes


cjpatton:
I'd rather not have to manually revert these changes, especially since it means I have to disable my linter and manually make the code uglier :)

Please land this and #97 ASAP so I can rebase my PR.

PR #99, opened 2018-06-24
Update client SCT list during TLS 1.3 hanshake, fixes #76

### LINKED ISSUE #76: Test ``TestSCTHandshake`` seems not to work with TLS 1.3
## Description
The test ``TestSCTHandshake`` requires 2 SignedCertificateTimestamps to be set.

## Expected behaviour
The test ``TestSCTHandshake`` requires 2 SignedCertificateTimestamps to be set on client's ConnectionState object after handshake is tested with a call to ``testHandshake()``.

## Current behaviour
Current behviour is as expected if TLS v1.2 (or earlier) is negotiated between parties. Nevertheless, if TLS v1.3 is negotiated   SignedCertificateTimestamps list is empty. This is causing following condition to be true and results in broken test
```
	actual := state.SignedCertificateTimestamps
	if len(actual) != len(expected) {
		t.Fatalf("got %d scts, want %d", len(actual), len(expected))
	}
```

, but this list is gets when run with TLS 1.3.

eklitzke:
I'm not 100% sure this is right, but I noticed #76 has an "open for community contribution" tag so I decided to take a stab. This fixes the test, anwyay.

It appears that the root issue is due to `hs.processServerHello()` updating various attributes on the client, among them the client SCTs. There's a different conditional taken for TLS 1.3 which doesn't invoke `hs.processServerHello()`, so with TLS 1.3 the client SCT list never gets updated.

The `hs.processServerHello()` method also sets `c.peerCertificates` and `c.verifiedChains` after doing some validation of `hs.session`. It's not clear to me exactly how this is supposed to work in TLS 1.3, but it's possible those attributes are also unset on the client in TLS 1.3.


henrydcase:
Perfect!! Thanks

handshake_client.go
```diff
@@ -251,6 +251,8 @@ func (hs *clientHandshakeState) handshake() error {
 		hs.keySchedule = newKeySchedule13(hs.suite, c.config, hs.hello.random)
 		hs.keySchedule.write(hs.hello.marshal())
 		hs.keySchedule.write(hs.serverHello.marshal())
+
+		c.scts = hs.serverHello.scts
```
henrydcase: I think this should be done in ``func (hs *clientHandshakeState) doTLS13Handshake() error {``. Do you mind moving it?

eklitzke:
Updated to move the logic to `doTLS13Handshake()`.

PR #106, opened 2018-07-13
Refactor the DC minting API

cjpatton:
This exports the `Credential` type and adds a new function `Delegate()` for creating a DC from a `*Credential`and a `*Certificate`.


henrydcase:
Closing the PR as per 
> "Hmm ... I may actually want to drop this PR"

PR #107, opened 2018-07-20
Remove the DC minting API

cjpatton:
This removes `NewDelegatedCredential` from the API, as well as the exported `DelegatedCredential` structure.The only dependency on this interface within tris were the tests; it was used to mint DCs for testing the extension in the handshake. These have been replaced with static test data: see `subcerts_test.go`. What's left here are the minimal changes required to be use the DC extension.

The reason for this change is that our needs for the minting API put a lot of pressure on tris. First, it's useful to have the ability to generate a DC key pair, then get the signature with another call. This would have meant changing tris' API. Second, we want to be able to mint DCs that use ed25519. Since ed25519 isn't available in mainline Go, this would have required vendering golang.org/x/crypto/ed25519.  Either of these changes make tris more difficult to maintain, especially if we want to be able to push these changes upstream.

Thus, the minting API, which we'll be consume by goKeyless (github.com/cloudflare.com/gokeylesss), will go elsewhere, perhaps in its own repository.

### DIFF
--- a/subcerts.go
+++ b/subcerts.go
@@ -5,29 +5,30 @@
 package tls
 
 // Delegated credentials for TLS
-// (https://tools.ietf.org/html/draft-ietf-tls-subcerts) is an IETF Internet
-// draft and proposed TLS extension. If the client supports this extension, then
-// the server may use a "delegated credential" as the signing key in the
-// handshake. A delegated credential is a short lived public/secret key pair
-// delegated to the server by an entity trusted by the client. This allows a
-// middlebox to terminate a TLS connection on behalf of the entity; for example,
-// this can be used to delegate TLS termination to a reverse proxy. Credentials
-// can't be revoked; in order to mitigate risk in case the middlebox is
-// compromised, the credential is only valid for a short time (days, hours, or
-// even minutes).
+// (https://tools.ietf.org/html/draft-ietf-tls-subcerts-01) is an IETF Internet
+// draft and proposed TLS extension. This allows a backend server to delegate
+// TLS termination to a trusted frontend. If the client supports this extension,
+// then the frontend may use a "delegated credential" as the signing key in the
+// handshake. A delegated credential is a short lived key pair delegated to the
+// server by an entity trusted by the client. Once issued, credentials can't be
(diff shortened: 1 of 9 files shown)


subcerts.go
```diff
@@ -83,11 +68,41 @@ func canDelegate(cert *x509.Certificate) bool {
 	return false
 }
 
-// This structure stores the public components of a credential.
+// credential stores the public components of a credential.
 type credential struct {
+	// The amount of time for which the credential is valid. Specifically, the
```
henrydcase: remove 'the'

subcerts.go
```diff
@@ -154,7 +170,10 @@ func unmarshalCredential(serialized []byte) (*credential, error) {
 // unmarshalSubjectPublicKeyInfo parses a DER encoded SubjectPublicKeyInfo
 // structure into a public key and its corresponding algorithm.
 func unmarshalSubjectPublicKeyInfo(serialized []byte) (crypto.PublicKey, SignatureScheme, error) {
-	publicKey, err := x509.ParsePKIXPublicKey(serialized)
+	var publicKey crypto.PublicKey
```
henrydcase: this change doesn't seem to bring anything

henrydcase:
LGTM; please address minor comments

subcerts.go
```diff
@@ -5,29 +5,30 @@
...
 //
-// BUG(cjpatton) Subcerts: Need to add support for PKCS1, PSS, and EdDSA.
-// Currently delegated credentials only support ECDSA. The delegator must also
-// use an ECDSA key.
+// This implements draft 01. This draft doesn't specify an object identifier for
+// the X.509 extension; we use one assigned by Cloudflare. In addition, IANA has
+// not assigned an extension ID for this extension; we picked up one that's not
+// yet taken.
+//
+// BUG(cjpatton) Only ECDSA is supported with delegated credentials for now;
```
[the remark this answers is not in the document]

subcerts.go

subcerts.go

Bren2010:
I don't really agree with this change, and I don't see the motivation for it. Don't block on my review.

PR #108, opened 2018-07-30
Update implementation of draft-ietf-tls-subcerts to draft 02

cjpatton:
This drops support for delegated credentials with TLS 1.2 and adds the protocol version and signature scheme to the credential structure.


henrydcase:
We need integration tests in order to move forward.

cjpatton:
@henrydcase, the build appears to be broken on master, not just my branch.

henrydcase:
@cjpatton nop, https://github.com/cloudflare/tls-tris/commit/77d1fbf2627abe447763f8698f826518994d7d4b. This commit is the last one and it was OK

cjpatton:
The output I just posted is with that commit checked out.

cjpatton:
Yup, so CI encountered a bug. All tests are passing now. I still need to figure out why test-bogo is failing on my machine.

bogus.txt
```diff
@@ -0,0 +1 @@
+This is added to test CI
```
henrydcase: ?

cjpatton:
For the record, I ran the following to get `make test` to work:
```
docker rmi tls-tris:bogo
docker rmi bogo-localserver
make -f _dev/Makefile build-all
cd _dev/bogo && docker build --no-cache .
```

bogus.txt
cjpatton: This was to make CI re-run. I'm removing the commit now.

henrydcase:
For the record - environment cleanup solves problems as described above.

subcerts.go
```diff
@@ -70,23 +70,26 @@ func canDelegate(cert *x509.Certificate) bool {
...
-	// credential expires `ValidTime` seconds after the `notBefore` of the
+	// the credential expires `ValidTime` seconds after the `notBefore` of the
 	// delegation certificate. The delegator shall not issue delegated
 	// credentials that are valid for more than 7 days from the current time.
 	//
 	// When this data structure is serialized, this value is converted to a
 	// uint32 representing the duration in seconds.
 	validTime time.Duration
 
+	// The signature scheme associated with the credential public key.
```
henrydcase: .. with the DC public key

subcerts.go
```diff
@@ -141,67 +148,71 @@ func (cred *credential) marshal() ([]byte, error) {
...
+	cred.raw = serialized
+	return serialized, nil
 }
 
 // unmarshalCredential decodes a credential and returns it.
 func unmarshalCredential(serialized []byte) (*credential, error) {
-	// Bytes 0-3 are the validity time field; bytes 4-6 are the length of the
-	// serialized SubjectPublicKeyInfo.
-	if len(serialized) < 6 {
+	paramsLen := 8
```
henrydcase: Is there a reason why paramsLen := 8  not 10?

If there is - comment why you do +2. If not make it 10

subcerts.go
```diff
@@ -141,67 +148,71 @@ func (cred *credential) marshal() ([]byte, error) {
...
-func unmarshalSubjectPublicKeyInfo(serialized []byte) (crypto.PublicKey, SignatureScheme, error) {
+// structure into a public key.
+func unmarshalSubjectPublicKeyInfo(serialized []byte) (crypto.PublicKey, error) {
 	publicKey, err := x509.ParsePKIXPublicKey(serialized)
 	if err != nil {
-		return nil, 0, err
+		return nil, err
 	}
 
 	switch pk := publicKey.(type) {
```
henrydcase: Can you remove this code bloat? I would prefer something like:
```
if publicKey.(type) == *ecdsa.PublicKey {
return publicKey.(type), nil
}

return nil, fmt.Errorf()
```
Also this function is not needed anymore. Also to be removed

subcerts.go
```diff
@@ -141,67 +148,71 @@ func (cred *credential) marshal() ([]byte, error) {
...
 // unmarshalCredential decodes a credential and returns it.
 func unmarshalCredential(serialized []byte) (*credential, error) {
-	// Bytes 0-3 are the validity time field; bytes 4-6 are the length of the
-	// serialized SubjectPublicKeyInfo.
-	if len(serialized) < 6 {
+	paramsLen := 8
+
+	// Bytes 0:4 are the valid_time field, 4:6 are the scheme field, 6:8 are the
+	// version field, and 8:10 are the length of the serialized
+	// SubjectPublicKeyInfo.
```
henrydcase: Bytes 10:... contain serialized SubjectPublicKeyInfo

subcerts.go
```diff
@@ -141,67 +148,71 @@ func (cred *credential) marshal() ([]byte, error) {
...
+		return nil, fmt.Errorf("unsupported delegation key type: %T", pk)
 	}
 }
 
 // getCredentialLen returns the number of bytes comprising the serialized
 // credential that starts at the beginning of the input slice. It returns an
 // error if the input is too short to contain a credential.
 func getCredentialLen(serialized []byte) (int, error) {
-	if len(serialized) < 6 {
+	paramsLen := 8
```
henrydcase: same here paramsLen should be 10

subcerts.go

subcerts.go
cjpatton: Done. The reason is that the parameters are 8 bytes; the +2 is for the length of the public key.

subcerts.go
cjpatton: What you suggest isn't syntactically correct, because `.(type)` can only happen in a `switch`:
```
GOROOT/linux_amd64/src/crypto/tls/subcerts.go:199:7: use of .(type) outside type switch
GOROOT/linux_amd64/src/crypto/tls/subcerts.go:199:18: type *ecdsa.PublicKey is not an expression
```
But I'm pretty sure this could be done with the `reflect` package. Want me to do that?

> Also this function is not needed anymore. Also to be removed
Which function?

subcerts.go
cjpatton: Removed this comment and clarified the message structure in `(cred *credential) marshal()`.

subcerts.go
cjpatton: Added comment explaining why.

subcerts.go
henrydcase: nop, google "type assertions in go". hint: we are using it a lot in tris

subcerts.go
```diff
@@ -140,68 +150,70 @@ func (cred *credential) marshal() ([]byte, error) {
...
+		expectedVersion:             version,
+		publicKey:                   pk,
+	}, nil
 }
 
 // unmarshalSubjectPublicKeyInfo parses a DER encoded SubjectPublicKeyInfo
-// structure into a public key and its corresponding algorithm.
-func unmarshalSubjectPublicKeyInfo(serialized []byte) (crypto.PublicKey, SignatureScheme, error) {
+// structure into a public key.
+func unmarshalSubjectPublicKeyInfo(serialized []byte) (crypto.PublicKey, error) {
```
henrydcase: this function,is the one I don't want

PR #111, opened 2018-08-07
Don't use VersionTLS13DraftXX as public API

henrydcase:
Moving to draft-28 break a TestDCHandshake test

```
--- FAIL: TestDCHandshake (0.03s)
	subcerts_test.go:303: test #2 (tls13) fails: delegated credential: signature invalid
	subcerts_test.go:309: test #2 (tls13) usedDC = false; expected true
```

### DIFF
--- a/subcerts_test.go
+++ b/subcerts_test.go
@@ -139,7 +139,7 @@ func init() {
 		Rand:         zeroSource{},
 		Certificates: nil,
 		MinVersion:   VersionTLS10,
-		MaxVersion:   VersionTLS13Draft22,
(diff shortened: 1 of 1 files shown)


henrydcase:
It was broken here:
https://travis-ci.org/cloudflare/tls-tris/jobs/413226081

cjpatton:
LGTM. Note that when you update to draft28 as the default, I'll need to regenerate the test data. Do you plan to drop the draft number altogether?

henrydcase:
I'll do it in 2 steps

PR #115, opened 2018-08-08
Minimal number of changes needed to udpate to draft-28

### LINKED ISSUE #109: draft-25: Add the header to additional data
TLS 1.3, draft-28 specifies AD as record header, i.e. 
```
additional_data = TLSCiphertext.opaque_type 
                            TLSCiphertext.legacy_record_version 
                            TLSCiphertext.length 
```
This will need to be implemented in tris in order to make it compatible with TLS 1.3

henrydcase:
```
    includes AD in authentication check of TLS records    
    * As per 5.2 of TLS 1.3 draft-28, the additional data is record header.

    tests: Update tests in order to support draft-28  
    * Interoperability: Updates NSS and BoringSSL versions to the
      one supporting draft-28
    * Bogo: Updates revision number to use tests for draft-28
    * FIX: makefile was using test-compat target instead of
      test-interop
```

Fixes #109


subcerts_test.go
```diff
@@ -252,7 +256,9 @@ var dcTesters = []struct {
 	name             string
 }{
 	{true, true, false, VersionTLS12, VersionTLS12, 0, "tls12", true, true, "tls12"},
-	{true, true, false, VersionTLS13, VersionTLS13, 0, "tls13", true, true, "tls13"},
+	// TODO: DC tests do not work with TLS 1.3
+	//       this is to be fixed by GH#108
+	// {true, true, false, VersionTLS13, VersionTLS13, 0, "tls13", true, true, "tls13"},
```
henrydcase: For Chris to provide with a fix

henrydcase:
*Not to be merged until DC is fixed*

subcerts_test.go
cjpatton: You will also need #108, since this PR drops support for DCs with TLS 1.2, as per draft-02 of the spec. The test cases have also changed.

henrydcase:
This is unrelated

On August 8, 2018 5:29:13 PM GMT+01:00, Christopher Patton <notifications@github.com> wrote:
>cjpatton commented on this pull request.
>
>
>
>> @@ -252,7 +256,9 @@ var dcTesters = []struct {
> 	name             string
> }{
>	{true, true, false, VersionTLS12, VersionTLS12, 0, "tls12", true,
>true, "tls12"},
>-	{true, true, false, VersionTLS13, VersionTLS13, 0, "tls13", true,
>true, "tls13"},
>+	// TODO: DC tests do not work with TLS 1.3
>+	//       this is to be fixed by GH#108
>+	// {true, true, false, VersionTLS13, VersionTLS13, 0, "tls13", true,
>true, "tls13"},
>
>You will also need #108, since this PR drops support for DCs with TLS
>1.2, as per draft-02 of the spec. The test cases have also changed.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/cloudflare/tls-tris/pull/115#discussion_r208649175

-- Wysłane za pomocą K-9 Mail.

subcerts_test.go
henrydcase: nop, I test data only

conn.go
```diff
@@ -358,6 +359,15 @@ func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert)
 				hc.additionalData[11] = byte(n >> 8)
 				hc.additionalData[12] = byte(n)
 				additionalData = hc.additionalData[:]
+			} else {
+				if len(payload) > int(uint16(1<<14)+256) {
```
Ko-: These typecasts are confusing. Everything is already an `int`. Maybe just  `if len(payload) > ((1 << 14) + 256) {`.

conn.go
```diff
@@ -358,6 +359,15 @@ func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert)
...
+					return false, 0, alertRecordOverflow
+				}
+				// Check AD header, see 5.2 of RFC8446
+				additionalData = make([]byte, 5)
+				additionalData[0] = byte(recordTypeApplicationData)
+				binary.BigEndian.PutUint16(additionalData[1:], VersionTLS12)
```
Ko-: In other places in the code 16-bit values are explicitly split using bitmasks. (E.g.: `hc.additionalData[11] = byte(payloadLen >> 8)`, `hc.additionalData[12] = byte(payloadlen)`.) While I think that using `encoding/binary` is more clear to the reader, it makes it inconsistent within this file.

conn.go
```diff
@@ -481,13 +491,17 @@ func (hc *halfConn) encrypt(b *block, explicitIVLen int) (bool, alert) {
...
 				// opaque type
 				payload = payload[:len(payload)+1]
 				payload[len(payload)-1] = b.data[0]
 				b.data[0] = byte(recordTypeApplicationData)
+
+				// Add AD header, see 5.2 of RFC8446
+				additionalData = make([]byte, 5)
+				additionalData[0] = byte(recordTypeApplicationData)
+				binary.BigEndian.PutUint16(additionalData[1:], VersionTLS12)
+				binary.BigEndian.PutUint16(additionalData[3:], uint16(payloadLen+overhead))
```
Ko-: Same thing as before.

conn.go
```diff
@@ -463,7 +473,7 @@ func (hc *halfConn) encrypt(b *block, explicitIVLen int) (bool, alert) {
 			payloadLen := len(b.data) - recordHeaderLen - explicitIVLen
 			overhead := c.Overhead()
 			if hc.version >= VersionTLS13 {
-				overhead++
+				overhead++ // TODO(kk): why this is done?
```
Ko-: I believe it's because of this: "The length (in bytes) of the following TLSCiphertext.encrypted_record, which is the sum of the lengths of the content and the padding, **plus one for the inner content type**, plus any expansion added by the AEAD algorithm."

conn.go
```diff
@@ -358,6 +359,15 @@ func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert)
...
 				additionalData = hc.additionalData[:]
+			} else {
+				if len(payload) > int((1<<14)+256) {
+					return false, 0, alertRecordOverflow
+				}
+				// Check AD header, see 5.2 of RFC8446
+				additionalData = make([]byte, 5)
+				additionalData[0] = byte(recordTypeApplicationData)
+				binary.BigEndian.PutUint16(additionalData[1:], VersionTLS12)
+				binary.BigEndian.PutUint16(additionalData[3:], uint16(len(payload)))
```
cjpatton: Payload is the encoded record, right? The payload should be equal to the content followed by a byte denoting the content type, followed by any number of 0 bytes.

Is padding permitted in our implementation?

conn.go
cjpatton: That's right.

conn.go
```diff
@@ -358,6 +359,15 @@ func (hc *halfConn) decrypt(b *block) (ok bool, prefixLen int, alertValue alert)
 				hc.additionalData[11] = byte(n >> 8)
 				hc.additionalData[12] = byte(n)
 				additionalData = hc.additionalData[:]
+			} else {
+				if len(payload) > int((1<<14)+256) {
+					return false, 0, alertRecordOverflow
+				}
+				// Check AD header, see 5.2 of RFC8446
+				additionalData = make([]byte, 5)
+				additionalData[0] = byte(recordTypeApplicationData)
```
cjpatton: It might be clearer to replace `recordTypeApplicationData` with `23` and `VersionTLS12` with `0x0303`, since the spec specifies these bytes concretely. Just a suggestion.

Here and below.

conn.go
```diff
@@ -463,7 +473,7 @@ func (hc *halfConn) encrypt(b *block, explicitIVLen int) (bool, alert) {
 			payloadLen := len(b.data) - recordHeaderLen - explicitIVLen
```
cjpatton: I would comment that, when this function is called and the connection is using 1.3,  `explicitIVLen` is guaranteed to be `0`.

conn.go
cjpatton: I wouldn't overload the word `overhead` here. This is supposed to be the tag length of the underlying AEAD, i.e., the output of `c.Overhead()`. I'd define a variable `recordLen`. If `hc.version >= VersionTLS13`, then let `recordLen = payloadLen + 1`. Then on  this line, replace `uint16(payloadLen+overhead))` with `uint16(recordLen+c.Overhead())`.

cjpatton:
This code looks correct to me; my comments address readability.

conn.go
henrydcase: so comment is needed

conn.go
henrydcase: ok, works for me

conn.go
henrydcase: done

conn.go
henrydcase: I think there is more refactoring than that needed for this entire function. this is coming in next PR.  This one I would like to keep minimal.

cjpatton: Sounds good.

PR #119, opened 2018-08-09
 Clients MUST check for supported_versions extensions and abort in case it was received on TLS version prior to 1.3

### LINKED ISSUE #118: Clients MUST check for supported_versions extensions and abort in case it was received on TLS version prior to 1.3
Current code
```
handshake_messages.go:
...
	svData := findExtension(data, extensionSupportedVersions)
	if svData != nil {
		if len(svData) != 2 {
			return alertDecodeError
		}
		if m.vers != VersionTLS12 {
			return alertDecodeError
		}
		m.vers = uint16(svData[0])<<8 | uint16(svData[1])
	}
```

As per 4.2.1, if svData is < 0x0304, client MUST abort with illegal_parameter (as server has negotiated TLS 1.2 or older, in which case it MUST NOT send back supportedVersions extension)

henrydcase:
Fixes #118

### DIFF
--- a/handshake_messages.go
+++ b/handshake_messages.go
@@ -1114,7 +1114,11 @@ func (m *serverHelloMsg) unmarshal(data []byte) alert {
 		if m.vers != VersionTLS12 {
 			return alertDecodeError
 		}
-		m.vers = uint16(svData[0])<<8 | uint16(svData[1])
+		rcvVer := binary.BigEndian.Uint16(svData[0:])
+		if rcvVer < VersionTLS13 {
+			return alertIllegalParameter
+		}
+		m.vers = rcvVer
 	}
 
 	for len(data) != 0 {


handshake_messages.go
```diff
@@ -1114,7 +1114,11 @@ func (m *serverHelloMsg) unmarshal(data []byte) alert {
 		if m.vers != VersionTLS12 {
 			return alertDecodeError
 		}
-		m.vers = uint16(svData[0])<<8 | uint16(svData[1])
+		rcvVer := binary.BigEndian.Uint16(svData[0:])
+		if rcvVer < VersionTLS13 {
```
cjpatton: The client should also check that `rcVers` is among the versions it offered in its "supported_version" extension. Is this check made? If not, add it; if it is, I'd move this check to wherever that is.

henrydcase:
Yes. This check is done.

On August 9, 2018 8:50:41 PM GMT+01:00, Christopher Patton <notifications@github.com> wrote:
>cjpatton commented on this pull request.
>
>Not sure this is the best place for this check. Sec. 4.2.1 also insists
>that the selected version was one that it offered. Is this check done?
>If so, this check should go there. 
>
>> If the “supported_versions” extension in the ServerHello contains a
>version not offered by the client or contains a version prior to TLS
>1.3, the client MUST abort the handshake with an “illegal_parameter”
>alert.
>
>> @@ -1114,7 +1114,11 @@ func (m *serverHelloMsg) unmarshal(data
>[]byte) alert {
> 		if m.vers != VersionTLS12 {
> 			return alertDecodeError
> 		}
>-		m.vers = uint16(svData[0])<<8 | uint16(svData[1])
>+		rcvVer := binary.BigEndian.Uint16(svData[0:])
>+		if rcvVer < VersionTLS13 {
>
>The client should also check that `rcVers` is among the versions it
>offered in its "supported_version" extension. Is this check made? If
>not, add it; if it is, I'd move this check to wherever that is.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/cloudflare/tls-tris/pull/119#pullrequestreview-144998241

-- Wysłane za pomocą K-9 Mail.

handshake_messages.go
henrydcase: it's done

PR #125, opened 2018-08-17
fix key share tests

marten-seemann:
Fix a few fields that should never be empty or zero.

Found by turning up `n` in https://github.com/cloudflare/tls-tris/blob/242e89da1b99db253bb40a501d3be847a0b09271/handshake_messages_test.go#L48.

### DIFF
--- a/handshake_messages_test.go
+++ b/handshake_messages_test.go
@@ -210,7 +210,7 @@ func (*serverHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
 	}
 
 	if rand.Intn(10) > 5 {
-		m.keyShare.group = CurveID(rand.Intn(30000))
+		m.keyShare.group = CurveID(rand.Intn(30000) + 1)
 		m.keyShare.data = randomBytes(rand.Intn(300)+1, rand)
 	}
 	if rand.Intn(10) > 5 {
--- a/handshake_messages_test.go
+++ b/handshake_messages_test.go
@@ -205,7 +205,7 @@ func (*serverHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
 		numSCTs := rand.Intn(4)
 		m.scts = make([][]byte, numSCTs)
 		for i := range m.scts {
-			m.scts[i] = randomBytes(rand.Intn(500), rand)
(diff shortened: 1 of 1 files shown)


handshake_messages_test.go
```diff
@@ -162,7 +162,7 @@ func (*clientHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
 	m.keyShares = make([]keyShare, rand.Intn(4))
 	for i := range m.keyShares {
 		m.keyShares[i].group = CurveID(rand.Intn(30000))
-		m.keyShares[i].data = randomBytes(rand.Intn(300), rand)
+		m.keyShares[i].data = randomBytes(rand.Intn(300)+1, rand)
```
henrydcase: This is OK as it doesn't make sense to have CurveID without any data in ``key_share``

handshake_messages_test.go
```diff
@@ -205,13 +205,13 @@ func (*serverHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
...
 		for i := range m.scts {
-			m.scts[i] = randomBytes(rand.Intn(500), rand)
+			m.scts[i] = randomBytes(rand.Intn(500)+1, rand)
 		}
 	}
 
 	if rand.Intn(10) > 5 {
-		m.keyShare.group = CurveID(rand.Intn(30000))
-		m.keyShare.data = randomBytes(rand.Intn(300), rand)
+		m.keyShare.group = CurveID(rand.Intn(30000) + 1)
```
henrydcase: as above

handshake_messages_test.go
```diff
@@ -205,13 +205,13 @@ func (*serverHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
 		numSCTs := rand.Intn(4)
 		m.scts = make([][]byte, numSCTs)
 		for i := range m.scts {
-			m.scts[i] = randomBytes(rand.Intn(500), rand)
+			m.scts[i] = randomBytes(rand.Intn(500)+1, rand)
```
henrydcase: OK, as per RFC 6962 (see point 3.3)

PR #136, opened 2018-09-16
fix: in TLSv1.3, the ALPN extension must be sent in EE

### LINKED ISSUE #126: Caddy-tris: ERR_SSL_VERSION_INTERFERENCE
I've built caddy with the tris crypto library of go and thus tried out to use tls1.3 with caddy.

I added/modified three lines to the caddytls/config.go: `"TLS1-3": TLS_CHACHA20_POLY1305_SHA256` as an additional cipher and to only use this one for TLS1.3 negotiation and `"tls1.3": tls.VersionTLS13` for the available TLS versions.

Additionally I stated in the same file, that the max version is tls1.3: 

```
if config.ProtocolMaxVersion == 0 {
                config.ProtocolMaxVersion = tls.VersionTLS13
        }
```
But still [dev.ssllabs.com/ssltest](https://dev.ssllabs.com/ssltest/analyze.html?d=h-neef.de&s=185.101.92.192&latest) says, that my server negotiates TLS1.3 with a blacklisted HTTP/2 ciphersuite, which is why probably chrome throws the error mentioned in the title.

Any ideas?

henrydcase:
serverHandshakeState::readClientHello was setting selected ALPN protocol
always on hs.hello.alpnProtocol, which is specific to TLS 1.2 and older.
Because of that server was marshalling ALPN to SH instead of EE.

Fixes #126

### DIFF
--- a/_dev/interop_test_runner
+++ b/_dev/interop_test_runner
@@ -15,6 +15,8 @@ RE_PATTERN_HELLO_TLS_13_RESUME   = "Hello TLS 1.3 \[resumed\] _o/"
 RE_PATTERN_HELLO_0RTT            = "^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT\] _o/$"
 # Checks if 0-RTT was used and confirmed
 RE_PATTERN_HELLO_0RTT_CONFIRMED  = "^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT confirmed\] _o/$"
+# ALPN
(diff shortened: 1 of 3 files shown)


Henrocker:
Great work! Going to test this right now.

henrydcase:
Perfect! Thanks @Henrocker for submitting original issue and for your help with testing it.

Henrocker:
Glad I could help! 

Additionally, updated the tlswg Implementations of TRIS and added my site to the public test servers: https://github.com/tlswg/tls13-spec/wiki/Implementations

PR #140, opened 2018-09-29
don't expect CertificateVerify when the client doesn't send any cert

marten-seemann:
Section 4.4.3 of the spec says

>  Clients MUST send this message whenever authenticating via a certificate (i.e., when the Certificate message is non-empty). When sent, this message MUST appear immediately after the Certificate message and immediately prior to the Finished message.

We're doing this correctly in the client, see https://github.com/cloudflare/tls-tris/blob/07ad1769c3d96ff6953d85b5d1e489389d0823ac/13.go#L915-L918
This PR fixes the server, such that it doesn't expect a CertificateVerify message if the certificate list in the Certificate message was empty.

### DIFF
--- a/13.go
+++ b/13.go
@@ -311,29 +311,31 @@ func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error 
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
Closing in favor of #142.

PR #142, opened 2018-09-29
check that the client sent a certificate, if requested

### LINKED ISSUE #141: Server resonds with incorrect error code when client sends empty list of certificates
When client authentication is requested by setting `ClientAuth`, the server requests a certificate from the client, but it neither verifies if the client actually sent a certificate (for `RequireAnyClientCert` and `RequireAndVerifyClientCert`).

marten-seemann:
Fixes #141.

This is the same code that is used for TLS <1.3 in https://github.com/cloudflare/tls-tris/blob/07ad1769c3d96ff6953d85b5d1e489389d0823ac/handshake_server.go#L581-L588


13.go
```diff
@@ -306,6 +306,14 @@ func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error
 
 		hs.keySchedule.write(certMsg.marshal())
 		certs := getCertsFromEntries(certMsg.certificates)
+		if len(certs) == 0 {
+			// The client didn't actually send a certificate
```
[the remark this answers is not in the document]
marten-seemann: It makes sense, but I thought we want to make as few changes as possible to the TLS 1.2 code. Is that right?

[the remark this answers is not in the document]
henrydcase: not anymore

13.go
```diff
@@ -349,11 +349,11 @@ func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error
 	}
 
 	// client authentication
-	if certMsg, ok := msg.(*certificateMsg13); ok {
```
henrydcase: this is still needed. Server needs to respond with an error in case client sends a certificate, but server didn't requested it
In 4.4.2:
"
   The client MUST send a Certificate message if and only if the server
   has requested client authentication
"
Same point was already raised here:
https://github.com/cloudflare/tls-tris/pull/69#discussion_r174789298

13.go
```diff
@@ -364,39 +364,37 @@ func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error
...
 		}
-		hs.keySchedule.write(certVerify.marshal())
 
 		// Read next chunk
 		msg, err = c.readHandshake()
 		if err != nil {
 			return err
 		}
-
-	} else if (c.config.ClientAuth >= RequestClientCert) && !c.didResume {
```
henrydcase: It was OK

13.go
```diff
@@ -364,39 +364,37 @@ func (hs *serverHandshakeState) readClientFinished13(hasConfirmLock bool) error
 			return err
 		}
 
-		// 4.4.3: CertificateVerify MUST appear immediately after Certificate msg
-		msg, err = c.readHandshake()
-		if err != nil {
-			return err
-		}
+		if len(certs) > 0 {
```
henrydcase: so this always true, isn't? hs.processCertsFromClient() is already checking if certificate has been sent and it will return an error if not

13.go
marten-seemann: This will still cause an `alertUnexpectedMessage`. If the server didn't request a cert, but the client sent it, the next message will be parsed as a Finished message, which will fail if it's a Certificate message.

13.go
marten-seemann: Not if `ClientAuth` is set to `VerifyClientCertIfGiven`.

13.go
marten-seemann: This was moved to the beginning of the `if` block.

henrydcase:
Thanks!

PR #145, opened 2018-10-15
Interoperability testing: SIDH

henrydcase:
* Patch to boringssl with support for SIDH/P503-X25519
  * Uses MSR implementation of SIDH
* Interoperability tests against tls-tris


henrydcase:
@Ko- Can you take a look?

Ko-:
Looks good to me! It's not really consistent with naming (`x22519sidh503`, `SIDH/P503-X25519` and `SIDH-P503-X25519`), but I guess that's a given by the naming schemes of NSS, BoringSSL and Tris. All interop tests pass here as well.

PR #153, opened 2019-02-21
 [sike] Refactor key agreement in TLS 1.3

henrydcase:
Previously there where two methods used for key agreemnt
tls.Conn::generateKeyShare and tls.Conn::deriveDHESecret. Both were
used on client and server side. Boolean flag is used in order to
differentiate between key agreement performed on client and on server
side. Which sucks badly.
In order to implement shared secret agreement with KEM it is better to
add method which implements server specific key agreement and provide
default implementation which reuses tls.Conn::generateKeyShare followed
by tls.Conn::deriveDHESecret.
Now, it is possible for most of the DH-style key agreements to reuse
default implementation and for KEM-style key agreement to provide server
specific implementation.

### DIFF
--- a/13.go
+++ b/13.go
@@ -61,8 +61,8 @@ type keySchedule13 struct {
 	config         *Config   // Used for KeyLogWriter callback, nil if keylogging is disabled.
 }
 
-// Interface implemented by DH key exchange strategies
-type dhKex interface {
+// Interface implemented by key exchange strategies
+type kex interface {
 	// c - context of current TLS handshake, groupId - ID of an algorithm
 	// (curve/field) being chosen for key agreement. Methods implmenting an
 	// interface always assume that provided groupId is correct.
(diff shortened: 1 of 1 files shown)


13.go
```diff
@@ -61,33 +61,64 @@ type keySchedule13 struct {
...
+		return nil, keyShare{}, err
+	}
+	return ss, publicKey, nil
 }
 
 // Key Exchange strategies per curve type
-type kexNist struct{}     // Used by NIST curves; P-256, P-384, P-512
-type kexX25519 struct{}   // Used by X25519
-type kexSIDHp503 struct{} // Used by SIDH/P503
+type kexNist struct{ defaultServerKEX }     // Used by NIST curves; P-256, P-384, P-512
```
alxdavids: Since SIDH is also an acronym and capitalised in the variable name, maybe this should be kexNIST?

13.go
```diff
@@ -232,14 +252,17 @@ CurvePreferenceLoop:
...
-
 	earlyClientCipher, _ := hs.keySchedule.prepareCipher(secretEarlyClient)
 
-	ecdheSecret := c.deriveDHESecret(ks, privateKey)
-	if ecdheSecret == nil {
-		c.sendAlert(alertIllegalParameter)
-		return errors.New("tls: bad ECDHE client share")
+	if ks.group == 0 {
+		c.sendAlert(alertInternalError)
+		return errors.New("tls: HelloRetryRequest not implemented") // TODO(filippo)
```
alxdavids: Ancient TODO?

13.go
```diff
@@ -601,19 +624,28 @@ func prepareDigitallySigned(hash crypto.Hash, context string, data []byte) []byt
...
 func (c *Conn) generateKeyShare(curveID CurveID) ([]byte, keyShare, error) {
-	if val, ok := dhKexStrat[curveID]; ok {
+	if val, ok := kexStrat[curveID]; ok {
 		return val.generate(c, curveID)
 	}
 	return nil, keyShare{}, errors.New("tls: preferredCurves includes unsupported curve")
 }
 
 // DH key agreement. ks stores public key, secretKey stores private key used for ephemeral
 // key agreement. Function returns shared secret in case of success or empty slice otherwise.
```
alxdavids: now returns error also

13.go
```diff
@@ -601,19 +624,28 @@ func prepareDigitallySigned(hash crypto.Hash, context string, data []byte) []byt
...
+func (c *Conn) keyAgreementClient(ks keyShare, secretKey []byte) ([]byte, error) {
+	if val, ok := kexStrat[ks.group]; ok {
+		return val.keyAgreementClient(c, ks, secretKey)
 	}
-	return nil
+	return nil, errors.New("tls: unsupported group")
+}
+
+// keyAgreementServer generates ephemeral keypair on the on the server side
+// and then uses 'keyShare' (client public key) to dervie shared secret
```
alxdavids: nit: dervie -> derive

13.go
```diff
@@ -989,10 +1021,10 @@ func (hs *clientHandshakeState) doTLS13Handshake() error {
 
 	// 0-RTT is not supported yet, so use an empty PSK.
 	hs.keySchedule.setSecret(nil)
-	ecdheSecret := c.deriveDHESecret(serverHello.keyShare, hs.privateKey)
-	if ecdheSecret == nil {
+	ecdheSecret, err := c.keyAgreementClient(serverHello.keyShare, hs.privateKey)
```
alxdavids: Nit: change variable name (to remove dhe)?

13.go
henrydcase: yeah, this was just copy paste from somewhere else. Indeed this is not implemented yet, which is a shame a bit

13.go
henrydcase: good idea

13.go
henrydcase: done

13.go
henrydcase: let's make it ``sharedSecret`` as that's what it is

13.go
henrydcase: comment above updated

PR #159, opened 2019-02-25
Adds X25519-SIKEp503 key agreement

henrydcase:
* Adds possibility to establish shared secret with hybrid X25519-SIKEp503
* SIKEp503 is a quantum resistant KEM, which produces 128-bit shared secret
* Details of the key exchange protocol are described here:
https://github.com/henrydcase/draft-kwiatkowski-tls-ecdhe-sike
(partially at the moment)
* Requires #153

### DIFF
--- a/13.go
+++ b/13.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	sidh "github_com/cloudflare/sidh/sidh"
+	sike "github_com/cloudflare/sidh/sike"
 	"golang_org/x/crypto/curve25519"
 )
 
@@ -32,14 +33,22 @@ const numSessionTickets = 2
 type secretLabel int
 
 const (
-	x25519SharedSecretSz = 32
+	// Both public key and shared secret size
+	x25519Sz = 32
 
-	P503PubKeySz                  = 378
-	P503PrvKeySz                  = 32
-	P503SharedSecretSz            = 126
-	SIDHp503Curve25519PubKeySz    = x25519SharedSecretSz + P503PubKeySz
-	SIDHp503Curve25519PrvKeySz    = x25519SharedSecretSz + P503PrvKeySz
-	SIDHp503Curve25519SharedKeySz = x25519SharedSecretSz + P503SharedSecretSz
+	SIDHp503PubKeySz              = 378
+	SIDHp503PrvKeySz              = 32
+	SIDHp503SharedSecretSz        = 126
+	SIDHp503Curve25519PubKeySz    = x25519Sz + SIDHp503PubKeySz
+	SIDHp503Curve25519PrvKeySz    = x25519Sz + SIDHp503PrvKeySz
+	SIDHp503Curve25519SharedKeySz = x25519Sz + SIDHp503SharedSecretSz
+
+	SIKEp503SharedSecretSz        = 16
+	SIKEp503CtSz                  = SIDHp503PubKeySz + 24
+	SIKEp503Curve25519CtSz        = x25519Sz + SIKEp503CtSz
+	SIKEp503Curve25519PubKeySz    = x25519Sz + SIDHp503PubKeySz
+	SIKEp503Curve25519PrvKeySz    = x25519Sz + SIDHp503PrvKeySz + 24
+	SIKEp503Curve25519SharedKeySz = x25519Sz + SIKEp503SharedSecretSz
 )
 
 const (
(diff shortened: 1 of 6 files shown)


13.go
```diff
@@ -1275,32 +1322,178 @@ func (kex *kexHybridSIDHp503X25519) generate(c *Conn, groupId CurveID) (private
...
+	// 'buf' is a concatenation of (private || public) key. I need public key
+	// when decapsulating in kexSIKEp503::keyAgreementClient.
+	buf := make([]byte, prvKey.Size()+pubKey.Size())
+	copy(buf, prvKey.Export())
+	copy(buf[prvKey.Size():], ks.data)
+	return buf, ks, nil
+}
+
+// keyAgreementClient performs KEM decapsulation. 'privateKey' is a concatenation
+// of (private || public) key
```
alxdavids: I don't understand this, it seems to me that ks.data should already contain the public key data? (based on line 1366)

_dev/tris-localserver/server.go
```diff
@@ -56,12 +56,12 @@ func NewServer() *server {
 }
 
 func enableQR(s *server, enableDefault bool) {
```
alxdavids: Nit: enableQR -> enablePQ for consistency

13.go
```diff
@@ -32,14 +33,22 @@ const numSessionTickets = 2
...
-	P503PrvKeySz                  = 32
-	P503SharedSecretSz            = 126
-	SIDHp503Curve25519PubKeySz    = x25519SharedSecretSz + P503PubKeySz
-	SIDHp503Curve25519PrvKeySz    = x25519SharedSecretSz + P503PrvKeySz
-	SIDHp503Curve25519SharedKeySz = x25519SharedSecretSz + P503SharedSecretSz
+	// Both public key and shared secret size
+	x25519Sz = 32
+
+	SIDHp503PubKeySz              = 378
+	SIDHp503PrvKeySz              = 32
```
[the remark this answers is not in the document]
henrydcase: SIKE and SIDH private key size is different. Not sure where in documentation it is but
```
     378 : Public Key
   +  32 : "real" random private key - as defied for SIDH (252 bits for Allice and 250 for Bob in case of P503)
   +  24 : Message (see comment below)
---------
      434
```

13.go
```diff
@@ -32,14 +33,22 @@ const numSessionTickets = 2
...
+	SIDHp503SharedSecretSz        = 126
+	SIDHp503Curve25519PubKeySz    = x25519Sz + SIDHp503PubKeySz
+	SIDHp503Curve25519PrvKeySz    = x25519Sz + SIDHp503PrvKeySz
+	SIDHp503Curve25519SharedKeySz = x25519Sz + SIDHp503SharedSecretSz
+
+	SIKEp503SharedSecretSz        = 16
+	SIKEp503CtSz                  = SIDHp503PubKeySz + 24
+	SIKEp503Curve25519CtSz        = x25519Sz + SIKEp503CtSz
+	SIKEp503Curve25519PubKeySz    = x25519Sz + SIDHp503PubKeySz
+	SIKEp503Curve25519PrvKeySz    = x25519Sz + SIDHp503PrvKeySz + 24
```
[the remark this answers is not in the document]
henrydcase: That's what makes private key of SIKE different than SIDH. So SIKE is using this modified "Hofheinz, Hovelmanns Kiltz" transform. This transform requires random value which is equal to level of security + 64 bits. For P503 it KEM shared secret size is 16 bytes, hence 16+8 = 24.

If you are interested in more details than "On the security of supersingular isogeny cryptosystems (S. Galbraith, et al., 2016, ePrint #859)" or you can look at SIKE spec and look at variable defined as "n" (1.4 and 4.3)

13.go
```diff
@@ -61,38 +70,76 @@ type keySchedule13 struct {
 	config         *Config   // Used for KeyLogWriter callback, nil if keylogging is disabled.
 }
 
-// Interface implemented by DH key exchange strategies
-type dhKex interface {
+// Interface implemented by key exchange strategies
+type kex interface {
```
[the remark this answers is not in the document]
henrydcase: that's correct

13.go
henrydcase: it's like that - to decapsulate I need private and public key of a client AND ciphertext received from the server. 'ks' contains ciphertext (or to put it differently ``ks`` contains keyShare information which comes form the server, in case of SIKE it is an output of private key encapsulation - ciphertext).

Now it should be clear why SIKE doc says that secret key has 434 bytes (this is amount of data that one needs to store in order to decapsulate).

henrydcase: I'll change variable name to ``theirsKS``. Maybe will make it more readable

PR #161, opened 2019-03-12
Fix: always store client hello in ConnectionState

henrydcase:
Fixes a bug. On a server side, ConnectionState.ClientHello was filled
    only in case server was handling TLSv1.3

### DIFF
--- a/handshake_server.go
+++ b/handshake_server.go
@@ -65,10 +65,9 @@ func (c *Conn) serverHandshake() error {
 	if err != nil {
 		return err
 	}
-
-	// For an overview of TLS handshaking, see https://tools.ietf.org/html/rfc5246#section-7.3
-	// and https://tools.ietf.org/html/draft-ietf-tls-tls13-18#section-2
+	c.clientHello = hs.clientHello.marshal()
 	c.buffering = true
+
 	if c.vers >= VersionTLS13 {
 		if err := hs.doTLS13Handshake(); err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


handshake_server.go
```diff
@@ -69,6 +69,7 @@ func (c *Conn) serverHandshake() error {
 	// For an overview of TLS handshaking, see https://tools.ietf.org/html/rfc5246#section-7.3
 	// and https://tools.ietf.org/html/draft-ietf-tls-tls13-18#section-2
 	c.buffering = true
+	c.clientHello = hs.clientHello.marshal()
```
Lekensteyn: What about moving this after line 67, just after successfully reading a CH?

henrydcase: yup, why not. I'll do that

handshake_server.go
```diff
@@ -69,6 +69,7 @@ func (c *Conn) serverHandshake() error {
 	// For an overview of TLS handshaking, see https://tools.ietf.org/html/rfc5246#section-7.3
```
henrydcase: Also this comment is dumb and outdated. obviously handshaking is described in some section of TLS 1.2 and 1.3 RFC

handshake_server.go
Lekensteyn: I would personally have kept this reference. The handshake is not trivial and having a quick overview of the steps would help others who are not TLS experts to understand this code and verify its correctness.

PR #163, opened 2019-03-13
tls13: remove TLS 1.3 specific default cipher settings

### LINKED ISSUE #162: BUG: default ciphersSuites for 1.3
Tris has a weird logic, if no TLSv1.3 ciphers are provided than default (v1.3 specific) are added, resulting in adding ciphers which users do not explicitly request. The logic should follow the rule - either user specifies ciphers to be used or it specifies nothing and then default ones are used.

```
func (c *Config) cipherSuites() []uint16 {
	s := c.CipherSuites
	if s == nil {
		s = defaultCipherSuites()
	} else if c.maxVersion() >= VersionTLS13 {
		// Ensure that TLS 1.3 suites are always present, but respect
		// the application cipher suite preferences.
		s13 := defaultTLS13CipherSuites()
		if !hasOverlappingCipherSuites(s, s13) {
			allSuites := make([]uint16, len(s13)+len(s))
			allSuites = append(allSuites, s13...)
			s = append(allSuites, s...)
		}
	}
	return s
}
```

Related bugs:
1. It can be assumed that caller has a good reason for not using default set of ciphers. For example, caller may specify not to use AES-128. Because of this bug - tris will allow it and it will allow it only for TLS 1.3. Contract with the caller is violated. 
2. Code (above) is creating an array full of zeros of length equal to ``len(s13)+len(s)`` and appends IDs from ``s13`` and ``s``. Returned array is equal to ``2*(len(s13)+len(s))`` - it means ``cipherSuites()`` may allow to use cipher suite with ID=0 (TLS_NULL_WITH_NULL_NULL) !? What for? Luckly it is not supported by tris... luckly.

henrydcase:
Fixes #162

### DIFF
--- a/common.go
+++ b/common.go
@@ -821,15 +821,6 @@ func (c *Config) cipherSuites() []uint16 {
 	s := c.CipherSuites
 	if s == nil {
 		s = defaultCipherSuites()
-	} else if c.maxVersion() >= VersionTLS13 {
-		// Ensure that TLS 1.3 suites are always present, but respect
-		// the application cipher suite preferences.
-		s13 := defaultTLS13CipherSuites()
-		if !hasOverlappingCipherSuites(s, s13) {
-			allSuites := make([]uint16, len(s13)+len(s))
-			allSuites = append(allSuites, s13...)
-			s = append(allSuites, s...)
-		}
 	}
 	return s
 }


henrydcase:
@Lekensteyn I'll ask for your opinion if it will be needed. This PR is not to be merged yet

henrydcase:
need to close it as I don't know how to delete a comment

henrydcase:
need to close it as I don't know how to delete a comment

Issue #165, opened 2019-04-05
Test against multiple Go releases


henrydcase:
Would be great to run tests the library against multiple versions of Go, for amd64, arm, arm64 and on linux, windows and macOS

Lekensteyn:
Right now it does not build on macOS due to issues in the Makefile with sed and cp, I'm working on that.

However it also seems that 1.12 moved some packages since chacha20poly1305 is no longer under a golang_org folder. Compare 1.12.1 on macOS:
/usr/local/Cellar/go/1.12.1/libexec/src/internal/x/crypto/chacha20poly1305
/usr/share/go-1.12/src/internal/x/crypto/chacha20poly1305 (golang-1.12-src in Debian sid)

to go 1.11.6:
/usr/share/go-1.11/src/vendor/golang_org/x/crypto/chacha20poly1305

Lekensteyn:
I have opened #166 to track the Go 1.12 issue. Go 1.13 has changed things again, possibly due to Go modules.

PR #167, opened 2019-04-09
Fix macOS build and add it to the test matrix

Lekensteyn:
* Options `-r` and `-R` are equivalent for GNU sed, but macOS requires
  `-R' or else it fails with:
  cp: the -H, -L, and -P options may not be specified with the -r option.

* Copying a directory with `-f` on macOS would replace an empty
  directory, resulting in vendor/henrydcase being created instead of
  vendor/github.com/henrydcase. Fix by copying `vendor/.` instead.

* macOS `sed -i` requires an additional argument while that would break
  GNU sed. Use perl instead which has consistent behavior. Once sidn and
  nobs are updated, the workaround can be removed from tris.

* travis: remove deprecated sudo key and add macOS to the test matrix.
___
Partially addresses #165 (it adds macOS support but does not test with different architectures like arm64). Only Go unittests are executed on macOS since Travis lacks support for Docker on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,16 @@
-sudo: required
 language: go
+
+matrix:
+  include:
+    - os: linux
+      env: TEST_SUITE=test-unit
+    - os: linux
+      env: TEST_SUITE=test-bogo
+    - os: linux
+      env: TEST_SUITE=test-interop
+    - os: osx
+      env: TEST_SUITE=test-unit
+  fast_finish: true
 
 services:
   - docker
@@ -7,19 +18,13 @@ services:
 go:
   - 1.11.x
 
-env:
-  - TEST_SUITE=test-unit
-  - TEST_SUITE=test-bogo
-  - TEST_SUITE=test-interop
-
-matrix:
-  fast_finish: true
-
 before_install:
   - make -f _dev/Makefile fmtcheck
 
 install:
-  - sudo pip install docker
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo pip install docker; fi
 
 script:
+  # Travis does not support Docker on macOS, disable it in the build-all target.
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export DOCKER='echo Unsupported - docker'; fi
   - make -f _dev/Makefile build-all && make -f _dev/Makefile "$TEST_SUITE"
--- a/_dev/Makefile
+++ b/_dev/Makefile
@@ -46,9 +46,10 @@ $(BUILD_DIR)/$(OS_ARCH)/.ok_$(VER_OS_ARCH): clean
 	mkdir -p "$(GOROOT_LOCAL)/pkg"
 
 # Copy src/tools from system GOROOT
-	cp -Hr $(GOROOT_ENV)/src $(GOROOT_LOCAL)/src
-	cp -Hr $(GOROOT_ENV)/pkg/include $(GOROOT_LOCAL)/pkg/include
-	cp -Hr $(GOROOT_ENV)/pkg/tool $(GOROOT_LOCAL)/pkg/tool
+	# macOS cp requires -R instead of -r. It makes no difference for GNU cp.
+	cp -HR $(GOROOT_ENV)/src $(GOROOT_LOCAL)/src
+	cp -HR $(GOROOT_ENV)/pkg/include $(GOROOT_LOCAL)/pkg/include
+	cp -HR $(GOROOT_ENV)/pkg/tool $(GOROOT_LOCAL)/pkg/tool
 
 # Swap TLS implementation
 	rm -r $(GOROOT_LOCAL)/src/crypto/tls/*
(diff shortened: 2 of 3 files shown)


Lekensteyn:
The bogo container does not seem to have perl...
```
perl -pi -e 's/sed -i/perl -pi -e/' /tmp/tmp.beMNng/nobs/Makefile
make[1]: perl: Command not found
make[1]: *** [/go/src/github.com/cloudflare/tls-tris/_dev/Makefile:55: /go/src/github.com/cloudflare/tls-tris/_dev/GOROOT/linux_amd64/.ok_go1.11.8_linux_amd64] Error 127
make[1]: Leaving directory '/go/src/github.com/henrydcase/crypto-tls-bogo-shim'
make: *** [Makefile:5: bin/crypto-tls-bogo-shim] Error 2
make: *** [test-bogo] Error 2
The command "make -f _dev/Makefile build-all && make -f _dev/Makefile "$TEST_SUITE"" exited with 2.
```

_dev/Makefile
```diff
@@ -46,9 +46,9 @@ $(BUILD_DIR)/$(OS_ARCH)/.ok_$(VER_OS_ARCH): clean
 	mkdir -p "$(GOROOT_LOCAL)/pkg"
 
 # Copy src/tools from system GOROOT
-	cp -Hr $(GOROOT_ENV)/src $(GOROOT_LOCAL)/src
```
henrydcase: why this change?

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
 
+matrix:
+  include:
+    - os: linux
+      env: TEST_SUITE=test-unit
+    - os: linux
+      env: TEST_SUITE=test-bogo
+    - os: linux
+      env: TEST_SUITE=test-interop
+    # Travis does not support Docker on macOS.
```
henrydcase: not needed here

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
+    - os: osx
+      env: TEST_SUITE=test-unit
+  fast_finish: true
+
 services:
   - docker
```
henrydcase: how this will work with mac?

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
 before_install:
   - make -f _dev/Makefile fmtcheck
 
 install:
-  - sudo pip install docker
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo pip install docker; fi
 
 script:
+  # Travis does not support Docker on macOS, disable building these.
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export DOCKER='echo Unsupported - docker'; fi
```
henrydcase: I don't think  this is needed as you are running "test-unit" target only

_dev/Makefile
Lekensteyn: See the first commit message:
* Options `-r` and `-R` are equivalent for GNU sed, but macOS requires
  `-R' or else it fails with:
  cp: the -H, -L, and -P options may not be specified with the -r option.

.travis.yml
Lekensteyn: It is ignored, from https://travis-ci.org/cloudflare/tls-tris/jobs/517919516:
> Services are not supported on osx

I considered adding it to each individual `os: linux` entry, but that would result in duplication which is why I opted to leave it here. I can move it to the matrix if you prefer that?

.travis.yml
Lekensteyn: Unfortunately it is needed due to build-all -> build-test-tris-client -> `$(DOCKER) build ...`:

https://github.com/cloudflare/tls-tris/blob/master/_dev/Makefile#L91-L93

Lekensteyn:
Thanks for your review.

> Ideally we would run all tests for macOS, but I think running only unit tests is good enough and surely much better than nothing. The important part is to ensure software builds correctly on mac. If unit tests are passing on linux, then there is a big chance they will pass on mac.

Indeed, it failed the build before on macOS due to lack of tests. That should be solved here. Since the docker tests run a Linux container, I hope not much is lost. It would not have worked anyway, running macOS binaries in Linux.

As for the bogo build failure, I could 1) add perl to the image, 2) revert to sed and special case macOS, or 3) use perl if available and assume GNU sed otherwise. I'll decide tomorrow.

henrydcase:
Definitely,
> 1) add perl to the image
Idea is good and hence perl is needed

_dev/Makefile
henrydcase: I know, but I would like this comment in the code (btw: I don't understand the commit message - it says something about "sed", while change is done to "cp")

.travis.yml
henrydcase: ok, I agree with you - this is correct, don't change

_dev/Makefile
Lekensteyn: I'll add the comment in the Makefile, but I think such details are not important to comment about (once it works, it will probably not be changed).

I'll fix the commit message, it should indeed have said cp. Thanks!

.travis.yml
Lekensteyn: I've added a comment to clarify this.

henrydcase:
Thanks!

Issue #158, opened 2019-02-22
tls library


henrydcase:
Let's make it separated library instead of being part of Go standard library.
If possible we should keep mechanism allowing to swap sources of tris with Go implementation

Lekensteyn:
Just for future readers, patching the standard Go library will still be needed until at least Go 1.12 depending on the use case. If you only need a TLS implementation with raw sockets, a standalone package might work after #169.

If you need support for HTTP2 using net/http, then it won't work because that package explicitly checks for an interface from `crypto/tls`.

With Go modules (which is enabled by default in 1.13?) it might be possible to replace certain libraries, but I have not fully investigated that possibility.

PR #169, opened 2019-04-23
Use standard import paths, preparing for Go modules

Lekensteyn:
Change import paths to use standard import paths, this is friendlier to
standard tooling like syntax checking. Additionally, `go mod vendor` and
`go get` now works without a custom Go compiler. Such import paths will
also be needed for Go 1.13.

Note that `go test` passes except for example_test.go which fails due to
importing crypto/tls instead of tris. Likewise for programs in _dev/.

When building with _dev/go.sh, it will still patch the import paths to
use internal import paths though. This is the default and won't change.
___
Partially addresses #158 by making it importable by default.

### DIFF
--- a/13.go
+++ b/13.go
@@ -21,9 +21,9 @@ import (
 	"sync/atomic"
 	"time"
 
-	sidh "github_com/cloudflare/sidh/sidh"
-	sike "github_com/cloudflare/sidh/sike"
-	"golang_org/x/crypto/curve25519"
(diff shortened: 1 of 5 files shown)


Lekensteyn:
This makes development using vim-go a nicer experience since autocompletion and lint now actually works (my main motivation for making this change now). In the future it might enable using Tris as a Go module, but that is a future task to verify (and might require changes to example_test.go).

henrydcase:
LGTM

PR #170, opened 2019-04-24
interop_test_runner: use Python 3

Lekensteyn:
'python2' does not exist on macOS, it is called 'python2.7'. Porting to
Python 3 is however more future-proof so do that instead. Rename the
file such that it can be run with pytest.

Do not listen on a fixed host port and use a random port instead.
Otherwise tests could fail if something is listening on those ports.
___
~~Note, with this modification, I noticed some ResourceWarnings after calling `exec_run` due to a Docker socket being leaked. These won't affect the test results though so I just left it there. See https://github.com/docker/docker-py/issues/1293#issuecomment-486250527~~ I removed the wait logic improvement due to the complexity and the fact that it triggers a bug in docker-py.

I upgraded to Xenial to ensure a newer python3-six version to solve a failure importing docker. https://github.com/docker/docker-py/issues/2294#issuecomment-486269556

With the file rename, I can now do run a subset of the tests during development:

    pytest _dev/interop_test_runner.py -k SIDH -v

and add `-nauto` to run those tests in parallel (requires `pytest-xdist`). This can complete all tests in 30 seconds (with 12 jobs). Another nice feature is the ability to pause when an assertion fails and inspect the context:

    pytest _dev/interop_test_runner.py -k 'test_SIDH and Server'  --pdb

### DIFF
--- a/_dev/Makefile
+++ b/_dev/Makefile
@@ -135,7 +135,7 @@ test-bogo:
 	$(DOCKER) run --rm -v $(PRJ_DIR):$(BOGO_DOCKER_TRIS_LOCATION) tls-tris:bogo
 
 test-interop:
-	$(DEV_DIR)/interop_test_runner -v
+	$(DEV_DIR)/interop_test_runner.py -v
 
 ###############
 #
--- a/_dev/interop_test_runner.py
+++ b/_dev/interop_test_runner.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import docker
 import unittest
@@ -8,13 +8,13 @@ import time
 # Regex patterns used for testing
 
 # Checks if TLS 1.3 was negotiated
-RE_PATTERN_HELLO_TLS_13_NORESUME = "^.*Hello TLS 1.3 \(draft .*\) _o/$|^.*Hello TLS 1.3 _o/$"
+RE_PATTERN_HELLO_TLS_13_NORESUME = r"^.*Hello TLS 1.3 \(draft .*\) _o/$|^.*Hello TLS 1.3 _o/$"
 # Checks if TLS 1.3 was resumed
-RE_PATTERN_HELLO_TLS_13_RESUME   = "Hello TLS 1.3 \[resumed\] _o/"
+RE_PATTERN_HELLO_TLS_13_RESUME   = r"Hello TLS 1.3 \[resumed\] _o/"
 # Checks if 0-RTT was used and NOT confirmed
-RE_PATTERN_HELLO_0RTT            = "^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT\] _o/$"
+RE_PATTERN_HELLO_0RTT            = r"^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT\] _o/$"
 # Checks if 0-RTT was used and confirmed
-RE_PATTERN_HELLO_0RTT_CONFIRMED  = "^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT confirmed\] _o/$"
+RE_PATTERN_HELLO_0RTT_CONFIRMED  = r"^.*Hello TLS 1.3 .*\[resumed\] \[0-RTT confirmed\] _o/$"
 # ALPN
 RE_PATTERN_ALPN = "ALPN protocol: npn_proto$"
 # Successful TLS establishement from TRIS
@@ -27,23 +27,24 @@ class Docker(object):
     def __init__(self):
         self.d = docker.from_env()
 
+    def close(self):
+        self.d.close()
+
     def get_ip(self, server):
         tris_localserver_container = self.d.containers.get(server)
         return tris_localserver_container.attrs['NetworkSettings']['IPAddress']
 
     def run_client(self, image_name, cmd):
         ''' Runs client and returns tuple (status_code, logs) '''
-        c = self.d.containers.create(image=image_name, command=cmd)
-        c.start()
+        c = self.d.containers.run(image=image_name, detach=True, command=cmd)
         res = c.wait()
-        ret = c.logs()
+        ret = c.logs().decode('utf8')
         c.remove()
         return (res['StatusCode'], ret)
 
     def run_server(self, image_name, cmd=None, ports=None, entrypoint=None):
         ''' Starts server and returns docker container '''
-        c = self.d.containers.create(image=image_name, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
-        c.start()
+        c = self.d.containers.run(image=image_name, auto_remove=True, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
         # TODO: maybe can be done better?
         time.sleep(3)
         return c
@@ -95,12 +96,16 @@ class InteropServer(object):
     @classmethod
     def setUpClass(self):
         self.d = Docker()
-        self.server = self.d.run_server(self.TRIS_SERVER_NAME)
+        try:
+            self.server = self.d.run_server(self.TRIS_SERVER_NAME)
+        except:
+            self.d.close()
+            raise
 
     @classmethod
     def tearDownClass(self):
         self.server.kill()
-        self.server.remove()
+        self.d.close()
 
     @property
     def server_ip(self):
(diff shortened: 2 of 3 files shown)


_dev/Makefile
```diff
@@ -135,7 +135,7 @@ test-bogo:
 	$(DOCKER) run --rm -v $(PRJ_DIR):$(BOGO_DOCKER_TRIS_LOCATION) tls-tris:bogo
 
 test-interop:
-	$(DEV_DIR)/interop_test_runner -v
+	$(DEV_DIR)/interop_test_runner.py -v
```
henrydcase: what for?

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
+        self.wait_for_ports(c)
         return c
 
+    def wait_for_ports(self, c):
+        ''' Waits until all services from the image are ready. '''
+        check_ports = set(int(portproto.split('/')[0]) for portproto in c.attrs['NetworkSettings']['Ports'])
+        for i in range(0, 100):
+            # 'ss -Htln' is the modern form, but boringssl image only includes
+            # netstat (via busybox).
+            info = c.exec_run('sh -c "ss -Htln 2>/dev/null || netstat -tln"')
```
henrydcase: use only ss or netstat. you can modify boringssl image

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
+            if info.exit_code != 0:
```
henrydcase: remove ``!= 0``

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
+                time.sleep(3)
+                return
+            tcp_listen_info = [line for line in info.output.decode('utf8').splitlines() if 'LISTEN' in line]
+            if tcp_listen_info:
+                tcp_ports = set(int(line.split()[3].split(':')[-1]) for line in tcp_listen_info)
+                check_ports -= tcp_ports
+            if not check_ports:
+                return
+            time.sleep(.1)
+        print('WARNING: services from %s are not ready, missing %s' % (c.image. check_ports))
```
henrydcase: Don't let the test start. Kill & remove server, return nil

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
-        c = self.d.containers.create(image=image_name, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
-        c.start()
-        # TODO: maybe can be done better?
-        time.sleep(3)
+        c = self.d.containers.run(image=image_name, auto_remove=True, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
+        self.wait_for_ports(c)
         return c
 
+    def wait_for_ports(self, c):
+        ''' Waits until all services from the image are ready. '''
```
henrydcase: that's not true

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
+                print('Unable to check for open ports, sleeping for a fixed time')
+                time.sleep(3)
+                return
+            tcp_listen_info = [line for line in info.output.decode('utf8').splitlines() if 'LISTEN' in line]
+            if tcp_listen_info:
+                tcp_ports = set(int(line.split()[3].split(':')[-1]) for line in tcp_listen_info)
+                check_ports -= tcp_ports
+            if not check_ports:
+                return
+            time.sleep(.1)
```
henrydcase: what for?

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
-        # TODO: maybe can be done better?
-        time.sleep(3)
+        c = self.d.containers.run(image=image_name, auto_remove=True, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
+        self.wait_for_ports(c)
         return c
 
+    def wait_for_ports(self, c):
+        ''' Waits until all services from the image are ready. '''
+        check_ports = set(int(portproto.split('/')[0]) for portproto in c.attrs['NetworkSettings']['Ports'])
+        for i in range(0, 100):
```
henrydcase: use ``_`` if you don't use ``i``

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
+            # 'ss -Htln' is the modern form, but boringssl image only includes
+            # netstat (via busybox).
+            info = c.exec_run('sh -c "ss -Htln 2>/dev/null || netstat -tln"')
+            if info.exit_code != 0:
+                print('Unable to check for open ports, sleeping for a fixed time')
+                time.sleep(3)
```
henrydcase: so you will sleep for 3 seconds and then return anyway?

_dev/interop_test_runner.py
```diff
@@ -27,27 +27,47 @@ class Docker(object):
...
         ''' Starts server and returns docker container '''
-        c = self.d.containers.create(image=image_name, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
-        c.start()
-        # TODO: maybe can be done better?
-        time.sleep(3)
+        c = self.d.containers.run(image=image_name, auto_remove=True, detach=True, command=cmd, ports=ports, entrypoint=entrypoint)
+        self.wait_for_ports(c)
         return c
 
+    def wait_for_ports(self, c):
```
henrydcase: can we do it better without calling shell at all?

_dev/Makefile
Lekensteyn: Otherwise `pytest` won't be able to load the script (see initial description for why pytest helps during development).

_dev/interop_test_runner.py
Lekensteyn: Ok, will ensure that ss is installed then.

_dev/interop_test_runner.py
Lekensteyn: Ok, it should be true when I apply your suggestion above to abort the test.

_dev/interop_test_runner.py
Lekensteyn: Yes, the old behavior. I'll just add a hard dependency on ss then and remove this general case.

_dev/interop_test_runner.py
Lekensteyn: I couldn't really find a good alternative. I tried to retrieve container IP and try to port scan it from Python. Does not work on Docker for Mac, [details](https://docs.docker.com/docker-for-mac/networking/#i-cannot-ping-my-containers).

An alternative is to start a container for the sole purpose of port-scanning a host (it could consist of a single Go binary), but the size of that support code will be slightly larger than the current approach. If you think that this would be better, I can change it though?

henrydcase: I would double check docker-py. I wonder if it's not done by this wrapper itself.
If not can't we just "import socket" and "socket.connect"? It's hard for me to  believe python doesn't have something equivalent to ``ss``

Lekensteyn: I have tried the socket approach before, but it won't work:
> Does not work on Docker for Mac, [details](https://docs.docker.com/docker-for-mac/networking/#i-cannot-ping-my-containers).

Checking for reachable ports is not something that can be done with the high-level docker-py API (https://github.com/docker/docker-py/issues/2229). Usually healthchecks are used to test whether a service is ready (e.g. aside from checking that a TCP socket can be opened, verify that the HTTP service returns a proper response). In our case, it is sufficient if a TCP socket can be opened.

henrydcase: ok, keep digging.

henrydcase:
This change is a root cause of python issuing such warn's
```
/usr/lib/python3.5/http/client.py:410: ResourceWarning: unclosed <socket.socket fd=8, family=AddressFamily.AF_UNIX, type=2049, proto=0, raddr=/var/run/docker.sock>
  self._close_conn()
```
I'm actually getting a lot of those locally. It will need to be fixed before merging

Lekensteyn:
I've dropped the whole server start optimization (and kept two versions at https://github.com/cloudflare/tls-tris/commits/pwu/tests-speedup-v0 and https://github.com/cloudflare/tls-tris/commits/pwu/tests-speedup-v1).

Now it only fixes the Python 3 issues, pytest support and some potential resource leaks.

PR #171, opened 2019-04-24
Remove dependency on tls.Conn in key agreement abstraction

Lekensteyn:
The key agreement protocols only need a source of randomness for key
generation and agreement (for KEM), and the role (client or server), so
pass these explicitly.

Make keyAgreementClient and keyAgreementServer independent of the
tls.Conn instance to allow reuse for ESNI.
___
In the TLS handshake, we have:
1. Client generates an ephemeral private key and sends a derived public key share.
2. Server takes this public key share, generates a new ephemeral private key, derives a shared secret and sends a derived public key share.
3. Client uses its earlier private key to compute the same shared secret.

For ESNI, we have:
1. Server generates a semi-static private key and publishes a derived public key share (say, in DNS).
2. Client generates a new ephemeral private key, derives a shared secret and sends a derived public key share.
3. Server uses its earlier private key to compute the shared secret.

As you can see the roles are reversed. It didn't seem correct to call keyAgreementClient as a server and vice versa. Neither am I sure if the "Alice" and "Bob" role should be reversed for SIDH. For ESNI I plan to support ECDHE only. ~~The `ecdheKex` interface is adapted for the TLS handshake (with PQ support), but for ESNI, this interface will directly be used.~~ The kex interface is maintained, instead keyAgreementXxx have been made independent of Conn. For ESNI I simply do not use the PQ algorithms, so the difference between keyAgreementClient/Server does not matter.

~~It is possible to specialize the "generate" method, for TLS it could receive "isClient" but in ECDHE there is no distinction so I could drop it. Thoughts?~~

### DIFF
--- a/13.go
+++ b/13.go
@@ -72,25 +72,21 @@ type keySchedule13 struct {
 
 // Interface implemented by key exchange strategies
 type kex interface {
-	// c - context of current TLS handshake, groupId - ID of an algorithm
-	// (curve/field) being chosen for key agreement. Methods implmenting an
-	// interface always assume that provided groupId is correct.
-	//
-	// In case of success, function returns secret key and ephemeral key. Otherwise
-	// error is set.
-	generate(c *Conn, groupId CurveID) ([]byte, keyShare, error)
-	// keyAgreementClient declares an API for implementing shared secret agreement on
-	// the client side. `c` is a context of current TLS handshake, `ks` is a public key
-	// received from the server, ``privateKey`` client private key.
-	// Function returns shared secret in case of success or non nil error otherwise.
-	keyAgreementClient(c *Conn, ks keyShare, privateKey []byte) ([]byte, error)
-	// keyAgreementServer declares an API for implementing shared secret agreement on
-	// the server side. `c` context of current TLS handshake, `ks` is a public key
-	// received from the client side of the connection, ``privateKey`` is a private key
-	// of a server.
-	// Function returns secret shared between parties and public value to exchange
-	// between parties. In case of failure `error` must be set.
-	keyAgreementServer(c *Conn, ks keyShare) ([]byte, keyShare, error)
+	// generate generates an ephemeral key. The caller must ensure that a
+	// valid curve or field is given. Returns a new ephemeral secret and
+	// public key on success and an error on failure.
+	generate(isClient bool, rand io.Reader, groupId CurveID) ([]byte, keyShare, error)
+
+	// keyAgreementClient derives a shared secret for the client given the
+	// server's public key share and the client's private key.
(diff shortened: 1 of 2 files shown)


Lekensteyn:
Regarding your suggestion to split generate into generateClient/generateServer, I have decided to do that for now since it would either result in some more code duplication or a special if/else case in kexHybridSIDHp503X25519.generate. This can be done later if desired.

As for renaming "Client" -> "Initiator" (the party that generates a key share first) and "Server" -> "Responder" (the party that receives the key share first before generating a new key share and shared secret), I think it might be more confusing for the TLS handshake so I have decided against it. I'll just add a comment to the ESNI code.

13.go
```diff
@@ -122,7 +118,6 @@ type kexX25519 struct{ defaultServerKEX }   // Used by X25519
 type kexSIDHp503 struct{ defaultServerKEX } // Used by SIDH/P503
 type kexSIKEp503 struct{}                   // Used by SIKE/P503
 type kexHybridSIDHp503X25519 struct {
-	defaultServerKEX
```
Lekensteyn: The default implementation was not sufficient as the client and server generates keys and a shared secret using different computations. Hence the specialization below.

13.go
```diff
@@ -1327,15 +1322,15 @@ func (kex *kexHybridSIDHp503X25519) generate(c *Conn, groupId CurveID) (private
 	return prvHybrid[:], keyShare{group: HybridSIDHp503Curve25519, data: pubHybrid[:]}, nil
 }
 
-func (kex *kexHybridSIDHp503X25519) keyAgreementClient(c *Conn, theirsKS keyShare, key []byte) ([]byte, error) {
+func (kex *kexHybridSIDHp503X25519) computeSharedSecret(isClient bool, theirsKS keyShare, key []byte) ([]byte, error) {
```
henrydcase: Why changing this?

13.go
```diff
@@ -1344,22 +1339,39 @@ func (kex *kexHybridSIDHp503X25519) keyAgreementClient(c *Conn, theirsKS keyShar
...
-	ret, err = kex.pqKEX.keyAgreementClient(c, tmpKs, key[x25519Sz:])
+	ret, err = kex.pqKEX.keyAgreementClient(isClient, tmpKs, key[x25519Sz:])
 	if err != nil {
 		return nil, err
 	}
 	copy(sharedKey[x25519Sz:], ret)
 	return sharedKey[:], nil
 }
 
+func (kex *kexHybridSIDHp503X25519) keyAgreementClient(theirsKS keyShare, key []byte) ([]byte, error) {
```
henrydcase: this is not needed (just change the name of function)

henrydcase:
LGTM

Issue #138, opened 2018-09-25
Support E-SNI


henrydcase:
The point of this ticket is to support E-SNI:

https://blog.cloudflare.com/encrypted-sni
https://tools.ietf.org/html/draft-ietf-tls-esni-01

f4nff:
good～

cetinsert:
@henrydcase @f4nff @agl @henrydcase @FiloSottile 
Is there a version of this library or any other Go TLS 1.3 library with E-SNI support available for testing?
I would like to setup a censorship circumvention tool for Korea.

f4nff:
There are currently no golang libraries that can be tested for e-sni.

Lekensteyn:
Status of the ecosystem:
- Latest version is https://tools.ietf.org/html/draft-ietf-tls-esni-03
- OpenSSL is waiting for the draft to be finished. https://github.com/openssl/openssl/issues/7482
- Firefox+NSS supports it https://bugzilla.mozilla.org/show_bug.cgi?id=1495120
- picotls supports it https://github.com/h2o/picotls/pull/155
- Go crypto/tls won't support it until ESNI is widely deployed: https://github.com/golang/go/issues/9671#issuecomment-439561672

Todo:
- Look at other implementations and try to come up with a reasonable API.

My initial thought is to include an array of [ESNIKeys](https://tools.ietf.org/html/draft-ietf-tls-esni-03#section-4.1) items. For TLS clients, the first item will be used. For TLS servers, a matching entry will be used. The DNS query should be handled outside crypto/tls, but we could provide some example code in a test (or maybe a separate package).

Alternatively, two functions fields could be created, `ClientESNIKeys: ESNIKeys` (which would be a single item) and `GetServerESNIKeys: func(record_digest []byte)`. The return values are TBD, but as the server can also provide ESNIKeys in its response for use with future connections, this must somehow be exposed as well.

I'll investigate this later.

PR #172, opened 2019-05-02
Add ESNI client and server support

Lekensteyn:
Implements https://tools.ietf.org/html/draft-ietf-tls-esni-01
Extends the tls.Config API with a ClientESNIKeys structure which must
contain a valid key. If this key is not valid, the handshake will fail.
A GetServerESNIKeys API is also added which allows the server to
dynamically query for an appropriate ESNI key.

Add a new 'esnitool' utility to generate ESNIKeys for testing purposes,
this uses a short lifetime, a single curve and cipher suite. The test
client and server can now be used with these keys. Additionally the test
client can securely query the ESNI key from DNS (hardcoded to use
1.1.1.1:853 using DoT for now).
___
Fixes https://github.com/cloudflare/tls-tris/issues/138

This change depends on https://github.com/cloudflare/tls-tris/pull/171 to ensure that the shared secret can be calculated without access to a `tls.Conn` structure. Therefore that commit is also included in this PR.


_dev/Makefile
```diff
@@ -99,18 +99,25 @@ build-all: \
...
 build-test-tris-client: $(BUILD_DIR)/$(OS_ARCH)/.ok_$(VER_OS_ARCH)
-	cd $(DEV_DIR)/tris-testclient; CGO_ENABLED=0 GOROOT="$(GOROOT_LOCAL)" $(GO) build -v -i .
+	cd $(DEV_DIR)/tris-testclient && GOROOT="$(GOROOT_LOCAL)" $(GO) get -v && \
+	   GOOS=linux CGO_ENABLED=0 GOROOT="$(GOROOT_LOCAL)" $(GO) build -v -i .
 	$(DOCKER) build -t tris-testclient $(DEV_DIR)/tris-testclient
 
 # Builds TRIS server
 build-test-tris-server: $(BUILD_DIR)/$(OS_ARCH)/.ok_$(VER_OS_ARCH)
-	cd $(DEV_DIR)/tris-localserver; CGO_ENABLED=0 GOROOT="$(GOROOT_LOCAL)" $(GO) build -v -i .
+	cd $(DEV_DIR)/tris-localserver; GOOS=linux CGO_ENABLED=0 GOROOT="$(GOROOT_LOCAL)" $(GO) build -v -i .
```
henrydcase: why?

handshake_messages.go
```diff
@@ -284,6 +290,15 @@ func (m *clientHelloMsg) marshal() []byte {
 		copy(z[5:], []byte(m.serverName))
 		z = z[l:]
 	}
+	if len(m.encryptedServerName) > 0 {
+		z[0] = byte(extensionEncryptedServerName >> 8)
```
henrydcase: use 	"encoding/binary" instead

handshake_messages.go
```diff
@@ -1296,6 +1320,15 @@ func (m *encryptedExtensionsMsg) marshal() []byte {
 		z = z[4:]
 	}
 
+	if len(m.esniNonce) > 0 {
+		z[0] = byte(extensionEncryptedServerName >> 8)
```
henrydcase: idem

_dev/Makefile
Lekensteyn: The binary that is built is included in a Docker image. Without this change, I had to run `GOOS=linux make -f _dev/Makefile all` in order to have working Docker images.

I can split this in a separate commit if you prefer that?

handshake_messages.go
Lekensteyn: Something like this?
```
l := len(m.encryptedServerName)
binary.BigEndian.PutInt16(z, extensionEncryptedServerName)
binary.BigEndian.PutInt16(z[2:], l)
copy(z[4:], m.encryptedServerName)
z = z[4+l:]
```

henrydcase: yup

Lekensteyn:
I've addressed the review comments, the GOOS=linux change was split in a separate commit. Since it the ESNI changes touch the same context, the ESNI patch effectively depends on that.

Aside from these two patches, there is also a Makefile patch to avoid unnecessary rebuilds. This is independent of ESNI, but included here for easier testing. I hope that PR #173 can be merged first. If it makes it easier for you, I could also drop that rebuild patch here and restrict PR #173 to that single change. Let me know!

Lekensteyn:
Ended up dropping the rebuild patch since it is not ready, will move that to PR #173. Now this patch is fully independent again and can be reviewed/merged independently.

henrydcase:
Can you improve tests?
At least handshake_messages_test.go, handshake_server_test.go and handshake_client_test.go should be augmented with some testing.

_dev/esnitool/esnitool.go
```diff
@@ -0,0 +1,137 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"time"
+
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/crypto/curve25519"
+)
+
+// Internal definitions, copied from common.go and esni.go
```
wbl: Might it make sense to have esni.go export them or equivalents?

common.go
```diff
@@ -641,6 +641,18 @@ type Config struct {
 	// This value has no meaning for the client.
 	GetDelegatedCredential func(*ClientHelloInfo, uint16) ([]byte, crypto.PrivateKey, error)
 
+	// ClientESNIKeys enables SNI encryption by the client using the
+	// provided server keys, it has no effect on a server. If the value is
+	// nil, then SNI encryption is not enabled.
+	//
+	// See https://tools.ietf.org/html/draft-ietf-tls-esni-01
+	ClientESNIKeys *ESNIKeys
```
wbl: This doesn't fit the pattern of a config being reused across connections since the ESNI keys are server specific. I can't think of a great way to do it, but one way would be to have a flag here to enable support, then DialWithDialer would use the resolver passed in to look up the text record.

tls_test.go
```diff
@@ -674,7 +674,7 @@ func TestCloneNonFuncFields(t *testing.T) {
 		switch fn := typ.Field(i).Name; fn {
 		case "Rand":
 			f.Set(reflect.ValueOf(io.Reader(os.Stdin)))
-		case "Time", "GetCertificate", "GetConfigForClient", "VerifyPeerCertificate", "GetClientCertificate", "GetDelegatedCredential":
+		case "Time", "GetCertificate", "GetConfigForClient", "VerifyPeerCertificate", "GetClientCertificate", "GetDelegatedCredential", "ClientESNIKeys", "GetServerESNIKeys":
```
wbl: ClientESNIKeys isn't a function

eighthave:
I'm working with a group, https://defo.ie/ on the ESNI standard.  There are some test servers there that might be useful.  Also, there are some tests in our openssl fork: https://github.com/sftcd/openssl

We also have working versions of nginx and lighttpd, and are building automated test suites to run in Travis CI and GitLab CI.

mvdan:
@Lekensteyn perhaps it would be a good idea to switch to a patch on top of the upcoming Go 1.14.x. It seems to contain all non-draft TLS 1.3 features, and it would smooth the way towards eventually incorporating it into upstream Go. I also imagine that it's best to get the latest fixes from upstream's crypto packages.

Lekensteyn:
@mvdan The plan to rebase has been an unwritten one, I have now opened #184 to track it.

PR #173, opened 2019-05-03
Makefile: avoid unnecessary rebuilds of the library

Lekensteyn:
* Do not wipe GOROOT for every `_dev/go.sh` invocation.
* Do not clone sidh and nobs repositories when only tris has changed.
* Do not copy and rebuild stdlib when only crypto/tls has changed.
* Avoid creating a temporary directory on every make invocation.
* Avoid creating src/src/ when the src/ directory already exists.
___
Important fix for integration of Tris in external projects which use the `_dev/go.sh` wrapper. WIthout this fix, it would rebuild Tris for every single `go` invocation.

### DIFF
--- a/_dev/Makefile
+++ b/_dev/Makefile
@@ -20,7 +20,6 @@ GOROOT_ENV  ?= $(shell $(GO) env GOROOT)
 GOROOT_LOCAL = $(BUILD_DIR)/$(OS_ARCH)
 # Flag indicates wheter invoke "go install -race std". Supported only on amd64 with CGO enabled
 INSTALL_RACE:= $(words $(filter $(ARCH)_$(shell go env CGO_ENABLED), amd64_1))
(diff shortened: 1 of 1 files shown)


Lekensteyn:
Changes since the initial PR:
* The bogo tests failed only on 1.12 because the bogo image was using 1.11 while the outside tests use 1.12.

Initially the Makefile depended on `$(GOROOT_LOCAL)/pkg` under the assumption that this directory will have an older mtime whenever `$(GOROOT_LOCAL)/pkg` is updated. This was not the case with the bogo test and prevented the `$(GOROOT_LOCAL)/src` from being updated to the latest version. As a fix, now the version is included (as `$(GOROOT_LOCAL)/pkg/.ok_$(VER_OS_ARCH)`.

Lekensteyn:
Hi @henrydcase, any comments on this patch? It would remove unnecessary git clones during the build and speed up development.

PR #176, opened 2019-08-05
Implement exporters for TLS 1.3

wbl:
This implements exporters for TLS 1.3.


wbl:
The tests are passing on my machine: I suspect Travis didn't pick up the commit as the hash looks different/maybe it's a go version problem. I think though I'll want to rework the API to bring it closer to upstream.

Lekensteyn:
The test-unit target is also failing on master. I'll have a look.

wbl:
Tests are passing now. This feature would come for free if we updated our starting point to a more recent upstream.

Lekensteyn:
What do you need this for? NTS?

I think it would be a good idea to not introduce an incompatible interface for the same functionality, upstream has ConnectionState.ExportKeyMaterial: https://golang.org/pkg/crypto/tls/#ConnectionState.ExportKeyingMaterial

We really should redo changes on top of upstream at some point.

PR #178, opened 2019-08-12
Makefile: fix vendoring issue for sike and sidh dependencies

### LINKED ISSUE #177: Vendoring issues with SIKE and SIDH
We currently rename import paths for github.com/henrydcase/nobs and github.com/cloudflare/sidh from github.com to github_com to avoid errors such as:
> can't load package: .../crypto/tls/13.go:24:2: non-standard import "github.com/cloudflare/sidh/sidh" in standard package "crypto/tls"

Since Go 1.13, non-standard import paths in `$GOROOT/src` are resolved with the vendor directory in `$GOROOT/src/vendor`, see https://github.com/golang/go/commit/fd080ea3bcc2, so this rename can be undone.

For Go 1.12 and before, we might have to find an alternative solution. Right now some projects that use `go build -mod=vendor` fail with:
```
make: Nothing to be done for 'go'.
.../_dev/GOROOT/linux_amd64/src/crypto/tls/13.go:24:2: cannot find package "." in:
	<other project>/vendor/github_com/cloudflare/sidh/sidh
.../_dev/GOROOT/linux_amd64/src/crypto/tls/13.go:25:2: cannot find package "." in:
	<other project>/vendor/github_com/cloudflare/sidh/sike
```

Addressing this issue potentially fixes #175 as well.

Lekensteyn:
Do not vendor github.com/cloudflare/sike and github.com/henrydcase/nobs
in Go 1.12 and before, that does not work when using -mod=vendor.
Instead use the 'internal/' prefix as is done for golang.org/x/ packages
in Go 1.12.

Go 1.13 makes it possible to use the vendor directory, so use that.

Fixes #177

### DIFF
--- a/_dev/Makefile
+++ b/_dev/Makefile
@@ -15,7 +15,7 @@ OS          ?= $(shell $(GO) env GOHOSTOS)
 ARCH        ?= $(shell $(GO) env GOHOSTARCH)
 OS_ARCH     := $(OS)_$(ARCH)
 VER_OS_ARCH := $(shell $(GO) version | cut -d' ' -f 3)_$(OS)_$(ARCH)
-VER_MAJOR   := $(shell $(GO) version | cut -d' ' -f 3 | cut -d. -f1-2)
+VER_MAJOR   := $(shell $(GO) version | cut -d' ' -f 3 | cut -d. -f1-2 | sed 's/[br].*//')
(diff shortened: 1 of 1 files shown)


Lekensteyn:
(rebased on #179, once that passes I'll merge this too)

PR #185, opened 2020-06-16
Fix building

claucece:
The location of 'https://github.com/henrydcase/crypto-tls-bogo-shim' changed to http://git.amongbytes.com/kris/crypto-tls-bogo-shim.git . I kept the location on the machine as '/go/src/github.com/kris/crypto-tls-bogo-shim' (github), as to not break other scripts.


claucece:
cc. @Lekensteyn

